PR #9, opened 2024-06-18
✨ add ability to run REST with grpc server

prashantgupta24:
## Description
This is temporary until [this](https://github.com/vllm-project/vllm/pull/5090#pullrequestreview-2113483815) review comment is addressed.

## Merge criteria:

- [ ] The commits are squashed in a cohesive manner and have meaningful messages.
- [ ] Testing instructions have been added in the PR body (for PRs involving changes that are not immediately obvious).
- [ ] The developer has manually tested the changes and verified that the changes work

PR #11, opened 2024-06-18
🐛 fix issue with starting both servers

prashantgupta24:
## Description

Pydantic didn't like `TYPE_CHECKING`:

```
pydantic.errors.PydanticUndefinedAnnotation: name 'ChatCompletionRequest' is not defined
```

Also had to remove unnecessary `yield` 

## Merge criteria:

### DIFF
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -6,15 +6,11 @@ import inspect
 import re
 from contextlib import asynccontextmanager
 from http import HTTPStatus
-from typing import TYPE_CHECKING
 
 import fastapi
 import uvicorn
 import vllm
-from fastapi import APIRouter
-
-if TYPE_CHECKING:
-    from fastapi import Request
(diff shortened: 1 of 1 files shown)


src/vllm_tgis_adapter/__main__.py
```diff
@@ -274,8 +266,6 @@ async def _force_log() -> None:
 
     server = await start_grpc_server(async_llm_engine, args)
 
-    yield
-
```
prashantgupta24, line 278: I'm assuming the `yield` was a no-op in this case?

dtrifiro: I think this won't work. see my comments on #9.

Perhaps something like this would work (although untested) https://gist.github.com/dtrifiro/251649959af43e3b5853bb8ab862d5a6

prashantgupta24:
closing in favor of https://github.com/opendatahub-io/vllm-tgis-adapter/pull/12

PR #20, opened 2024-06-26
Respect trace headers in grpc server (IBM #49)

prashantgupta24:
For https://github.com/IBM/vllm/pull/49

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,12 +28,17 @@ dynamic = ["version"]
 dependencies = [
   "vllm>=0.5.0",
   "prometheus_client==0.20.0",
-  "grpcio==1.64.1",
-  "grpcio-health-checking==1.62.1",
-  "grpcio-reflection==1.64.1",
+  "grpcio==1.62.2",
+  "grpcio-health-checking==1.62.2",
+  "grpcio-reflection==1.62.2",
   "transformers",
   "accelerate==0.31.0",
-  "hf-transfer==0.1.6"
+  "hf-transfer==0.1.6",
+  # additional dependencies for OpenTelemetry tracing
+  "opentelemetry-sdk",
+  "opentelemetry-api",
+  "opentelemetry-exporter-otlp",
+  "opentelemetry-semantic-conventions-ai"
 ]
 
 [project.urls]
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -51,6 +51,18 @@ from .pb.generation_pb2 import (
 )
 from .validation import validate_input, validate_params
 
+try:
+    from vllm.tracing import (
+        contains_trace_headers,
+        extract_trace_headers,
+        log_tracing_disabled_warning,
+    )
+except ImportError:
+    _vllm_tracing_available = False
+else:
+    _vllm_tracing_available = True
+
+
 if TYPE_CHECKING:
     import argparse
     from collections.abc import AsyncIterator, MutableSequence
@@ -224,12 +236,21 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
                 prompt=req.text,
                 prompt_token_ids=input_ids,
             )
+            kwargs = {}
+            if _vllm_tracing_available:
+                is_tracing_enabled = await self.engine.is_tracing_enabled()
+                headers = dict(context.invocation_metadata())
+                if is_tracing_enabled:
+                    kwargs["trace_headers"] = extract_trace_headers(headers)
+                elif contains_trace_headers(headers):
+                    log_tracing_disabled_warning()
             generators.append(
                 self.engine.generate(
                     inputs=inputs,
                     sampling_params=sampling_params,
                     request_id=f"{request_id}-{i}",
                     **adapter_kwargs,
+                    **kwargs,
                 ),
             )
 


pyproject.toml
```diff
@@ -30,10 +30,15 @@ dependencies = [
...
+  "grpcio-reflection==1.62.2",
   "transformers",
   "accelerate==0.31.0",
-  "hf-transfer==0.1.6"
+  "hf-transfer==0.1.6",
+  # additional dependencies for OpenTelemetry tracing
+  "opentelemetry-sdk",
+  "opentelemetry-api",
+  "opentelemetry-exporter-otlp",
+  "opentelemetry-semantic-conventions-ai"
```
prashantgupta24, line 41: ideally they should be in a separate section?

pyproject.toml
```diff
@@ -30,10 +30,15 @@ dependencies = [
   "prometheus_client==0.20.0",
   "grpcio==1.64.1",
   "grpcio-health-checking==1.62.1",
-  "grpcio-reflection==1.64.1",
+  "grpcio-reflection==1.62.2",
```
prashantgupta24, line 33: This is needed unfortunately because of version conflicts

prashantgupta24:
Not sure if the CI will work because for the `vllm-tgis-adapter` to be able to use `tracing.py`, it needs a `vllm` release that has that change? The PR is merged to main but no release of `vllm` has happened yet that has that change

pyproject.toml
dtrifiro: Does setting all grpc packages version to `1.64.1` work?

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -224,11 +229,19 @@ async def Generate(
                 prompt=req.text,
                 prompt_token_ids=input_ids,
             )
+            is_tracing_enabled = await self.engine.is_tracing_enabled()
+            headers = dict(context.invocation_metadata())
+            trace_headers = None
+            if is_tracing_enabled:
+                trace_headers = extract_trace_headers(headers)
+            if not is_tracing_enabled and contains_trace_headers(headers):
+                log_tracing_disabled_warning()
```
dtrifiro, line 238:
```suggestion
            if is_tracing_enabled:
                trace_headers = extract_trace_headers(headers)
            elif contains_trace_headers(headers):
                log_tracing_disabled_warning()
```

pyproject.toml
dtrifiro: These are not used in this project, are these actually required?

prashantgupta24: Huh, good point. Perhaps they are needed when you start vllm with `--otlp-traces-endpoint` enabled? let me check

prashantgupta24: Yeah look at this: https://github.com/opendatahub-io/vllm/blob/main/vllm/config.py#L1404-L1407

pyproject.toml
dtrifiro: Maybe it's worth it set all grpc packages to the same version, just to be on the safe side

prashantgupta24: Let me try that

prashantgupta24: Trying with 
```
(vllm) pip3 list | grep grpc
grpcio                                   1.62.2
grpcio-health-checking                   1.62.2
grpcio-reflection                        1.62.2
```

prashantgupta24: Seems to work!

dtrifiro:
So, the problem with CI is that the tracing part in vLLM hasn't been released yet upstream. We do have it in our image so this would be safe to merge, but CI will always fail till it is released upstream.

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -20,6 +20,11 @@
 from vllm.engine.async_llm_engine import _AsyncLLMEngine
 from vllm.entrypoints.openai.serving_completion import merge_async_iterators
 from vllm.inputs import TextTokensPrompt
+from vllm.tracing import (
+    contains_trace_headers,
+    extract_trace_headers,
+    log_tracing_disabled_warning,
+)
```
dtrifiro, line 27:
```suggestion
try:
    from vllm.tracing import (
        contains_trace_headers,
        extract_trace_headers,
        log_tracing_disabled_warning,
)
except ImportError:
    _vllm_tracing_available = False
else:
    _vllm_tracing_available = True
```

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -224,11 +229,19 @@ async def Generate(
                 prompt=req.text,
                 prompt_token_ids=input_ids,
             )
+            is_tracing_enabled = await self.engine.is_tracing_enabled()
+            headers = dict(context.invocation_metadata())
+            trace_headers = None
+            if is_tracing_enabled:
+                trace_headers = extract_trace_headers(headers)
+            elif contains_trace_headers(headers):
+                log_tracing_disabled_warning()
```
dtrifiro, line 238:
```suggestion
            if _vllm_tracing_available:
                is_tracing_enabled = await self.engine.is_tracing_enabled()
                headers = dict(context.invocation_metadata())
                trace_headers = None
                if is_tracing_enabled:
                    trace_headers = extract_trace_headers(headers)
                elif contains_trace_headers(headers):
                    log_tracing_disabled_warning()
```

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -224,11 +229,19 @@ async def Generate(
...
             generators.append(
                 self.engine.generate(
                     inputs=inputs,
                     sampling_params=sampling_params,
                     request_id=f"{request_id}-{i}",
+                    trace_headers=trace_headers,
```
dtrifiro, line 244:
```python
kwargs = {}
if _vllm_tracing_enabled:
    kwargs['trace_headers'] = trace_headers
```

```suggestion
                    **kwargs,
```

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: Something like this might be enough until a new version is released.

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -224,11 +234,20 @@ async def Generate(
...
                 prompt_token_ids=input_ids,
             )
+            if _vllm_tracing_available:
+                is_tracing_enabled = await self.engine.is_tracing_enabled()
+                headers = dict(context.invocation_metadata())
+                trace_headers = None
+                if is_tracing_enabled:
+                    trace_headers = extract_trace_headers(headers)
+                elif contains_trace_headers(headers):
+                    log_tracing_disabled_warning()
```
dtrifiro, line 244: missing the kwargs definition here

prashantgupta24: oh whoops, GH review comment missed that one

dtrifiro (approved):

PR #30, opened 2024-07-02
gha: test multiple versions

dtrifiro:
- gha: test multiple vLLM versions
- tests: use real vLLM version (stop using mocks)
- grpc_server: only print memory summary for CUDA devices
- tests: add real tests (for both grpc and http server)
- enable code coverage

The reason for this:
1. We can test multiple vLLM version and catch any issues before they hit PyPi
2. We can test a real vLLM engine by using the CPU executor

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -79,9 +79,3 @@ jobs:
 
       - name: Build package
         run: nox -s build-${{ matrix.pyv }}
-
-      - name: Upload artifact
-        uses: actions/upload-artifact@v4
-        with:
-          name: vllm-tgis-wheel
-          path: dist/vllm_tgis_adapter*.whl
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -11,6 +11,11 @@ on:
 
 env:
   FORCE_COLOR: "1"
+  # facilitate testing by building vLLM for CPU when needed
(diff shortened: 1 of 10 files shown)


dtrifiro:
Still missing: streaming tests

.github/workflows/tests.yaml
```diff
@@ -25,6 +30,11 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         pyv: ["3.11"]
+        vllm_version:
+          # - "pypi" # skip the pypi version as it will not work with CPU
```
dtrifiro, line 34: We cannot run the pypi version because it will try to use cuda even if `--device=cpu` is passed. This is due to how `vllm` is built and internally checks for cpu support (i.e by checking if `cpu` is in the vllm version.

dtrifiro: Luckily, by providing `VLLM_TARGET_DEVICE` and not building the avx512 version, we can quickly build the wheel from source .

prashantgupta24:
https://github.com/opendatahub-io/vllm-tgis-adapter/blob/main/src/vllm_tgis_adapter/__main__.py#L251

`image_input_type` was removed here - https://github.com/vllm-project/vllm/pull/5852

dtrifiro:
Handling image input shape issue in #31

prashantgupta24 (approved):

PR #40, opened 2024-07-12
Fix use of LoRA tokenizers

njhill:
This depends on vLLM PR https://github.com/vllm-project/vllm/pull/6227, which is only held up because I need to find time to add a test to it.

### DIFF
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 
 import fastapi
 import vllm
+from fastapi import APIRouter
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, Response, StreamingResponse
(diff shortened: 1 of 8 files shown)


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -706,39 +718,36 @@ def _convert_tokens(  # noqa: PLR0913
...
-        input_ids = await self.tokenizer_group.encode_async(
-            prompt,
-            **tokenize_kwargs,
-        )
 
-        # TODO this is temporary until truncation option is added
-        # to the TokenizerGroup encode methods
-        if truncate_input_tokens and truncate_input_tokens < len(input_ids):
-            input_ids = input_ids[-truncate_input_tokens:]
-            if not sampling_params.skip_special_tokens:
```
njhill, line 734: This was actually a bug, it shouldn't have been checking this param here (it applies to output tokens not input tokens).

dtrifiro:
Merge after #47

src/vllm_tgis_adapter/__main__.py
```diff
@@ -45,18 +50,30 @@
...
+
+try:
+    from vllm.entrypoints.openai.serving_tokenization import (
+        OpenAIServingTokenization,  # noqa: TCH002
+    )
+except ImportError:  #  vllm<=0.5.2
+    has_tokenization = False
+else:
+    has_tokenization = True
+
```
dtrifiro, line 62: This avoids lora breaking for <=0.5.2

tests/conftest.py
```diff
@@ -36,23 +37,44 @@ def monkeysession():
...
+def lora_adapter_name():
+    return "lora-test"
+
+
+@pytest.fixture()
+def lora_adapter_path():
+    from huggingface_hub import snapshot_download
+
+    path = snapshot_download(repo_id="yard1/llama-2-7b-sql-lora-test")
+    return path
```
dtrifiro, line 50: This is the same adapter used in the docs: https://docs.vllm.ai/en/latest/models/lora.html

tests/test_grpc_server.py
```diff
@@ -48,3 +49,10 @@ def test_batched_generation_request(grpc_client, grpc_server_thread_port):
 
     assert len(responses) == 2
     assert all(response.text for response in responses)
+
+
+@pytest.mark.skipif(vllm.config.is_cpu(), reason="Lora is not supported on cpu")
+def test_lora_request(grpc_client, lora_adapter_name):
+    response = grpc_client.make_request("hello", adapter_id=lora_adapter_name)
+
+    assert response.text
```
dtrifiro, line 58: @njhill just a simple sanity check for the grpc API, is there anything else you would like to test?

prashantgupta24 (approved):

PR #46, opened 2024-07-18
update for 0.5.2

dtrifiro:
- tgis_utils: base EnvVarArgumentParser on FlexibleArgumentParser
- update `__main__` for vllm>=0.5.2
- gha: bump tested version to 0.5.2

fixes RHOAIENG-10066 RHOAIENG-10063

### DIFF
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -11,6 +11,7 @@ from typing import TYPE_CHECKING
 
 import fastapi
 import vllm
+from fastapi import APIRouter
 from fastapi.exceptions import RequestValidationError
 from fastapi.middleware.cors import CORSMiddleware
 from fastapi.responses import JSONResponse, Response, StreamingResponse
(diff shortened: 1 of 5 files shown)


src/vllm_tgis_adapter/__main__.py
```diff
@@ -45,18 +50,30 @@
 
     from vllm.config import ModelConfig
 
+
+try:
```
dtrifiro, line 54: Tokenization endpoints were introduced after 0.5.2 was released, leaving this here to prepare for next releases

src/vllm_tgis_adapter/__main__.py
```diff
@@ -72,15 +89,42 @@ async def health() -> Response:
...
 async def show_available_models() -> JSONResponse:
-    models = await openai_serving_chat.show_available_models()
+    models = await openai_serving_completion.show_available_models()
     return JSONResponse(content=models.model_dump())
 
 
 @router.get("/version")
 async def show_version() -> JSONResponse:
-    ver = {"version": vllm.__version__}
+    ver = {"version": vllm.__version__, "commit": vllm.__commit__}
```
dtrifiro, line 127: This becomes slightly different than the upstream code, by also including the commit hash

prashantgupta24 (approved):

src/vllm_tgis_adapter/__main__.py
njhill: @dtrifiro the tokenization & detokenization endpoints were introduced in 0.5.1?

src/vllm_tgis_adapter/__main__.py
```diff
@@ -89,11 +133,15 @@ async def create_chat_completion(
     request: ChatCompletionRequest,
     raw_request: fastapi.Request,
 ) -> JSONResponse:
-    generator = await openai_serving_chat.create_chat_completion(request, raw_request)
+    generator = await openai_serving_chat.create_chat_completion(
+        request,
+        raw_request,
+    )
```
njhill, line 139: It would be good to keep the formatting the same as the upstream `openai_server.py` that this is a copy of, so that we can diff them easily.

I guess we are hoping to get rid of the duplication soon anyhow? In which case I guess it doesn't matter so much

PR #49, opened 2024-07-18
⚗️ try to patch prompt loads

prashantgupta24:
Upstream PR here: https://github.com/vllm-project/vllm/pull/6709

The motive to move upstream is because the monkey patching does not work in sharded mode.

Joe:
> I think the patch would only work if:
    - the patching happens on every shard, or
    - the patching happens before the other shards are booted, and fork is used instead of spawn, to preserve the patches

Update:
Decision made with Neil and team that we'll go ahead with monkey-patching with known issue of not working with TP.


src/vllm_tgis_adapter/tgis_utils/monkey_patch.py
```diff
@@ -0,0 +1,37 @@
...
+def from_local_checkpoint(
+    cls: "PromptAdapterModel",
+    adapter_path: str,
+    prompt_adapter_id: int,
+    torch_device: str = "cuda",
+) -> "PromptAdapterModel":
+    """Patch for PromptAdapterModel that supports decoder.pt models."""
+    peft_config_path = Path(adapter_path) / "adapter_config.json"
+    decoder_pt_path = Path(adapter_path) / "decoder.pt"
+
```
prashantgupta24, line 20: There's also this extra check in upstream:
```
 if num_virtual_tokens > config.max_prompt_adapter_token:
            raise ValueError(
                f'num_virtual_tokens ({num_virtual_tokens}) should be <= '
                f'max_prompt_adapter_token({config.max_prompt_adapter_token})')
```

dtrifiro:
Can we add a test?

prashantgupta24:
@joerunde is it worth addressing the upstream comment https://github.com/vllm-project/vllm/pull/6709#issuecomment-2246452657 if this PR won't get merged?

prashantgupta24:
@dtrifiro writing a unit test is proving harder than I anticipated, the adapters are not set up to use directly from HF since HF downloads them to `snapshots`, but the adapter code here assumes the `adapter_config.json` lives directly under the model name. 

I've added a bunch of integration tests for it though

prashantgupta24:
I can manually add a prompt adapter model and check it in, that works well with the unit tests. Only thing is that it's for `bloomz-560m`, I couldn't find one for the default `facebook-opt125m`

tests/conftest.py
```diff
@@ -40,8 +44,8 @@ def monkeysession():
 
 @pytest.fixture(scope="session")
 def lora_available() -> bool:
-    # lora does not work on cpu
-    return not vllm.config.is_cpu()
+    # lora does not work with the current set of models
+    return False
```
prashantgupta24, line 48: I had to disable lora explicitly since I got these errors
```
Model OPTForCausalLM does not support LoRA, but LoRA is enabled

Model BloomForCausalLM does not support LoRA, but LoRA is enabled
```

dtrifiro:
> I can manually add a prompt adapter model and check it in, that works well with the unit tests. Only thing is that it's for bloomz-560m, I couldn't find one for the default facebook-opt/125m

How large is that? We can add it as a tests which runs only with specific pytest flags (`--longtests` or something like that)

prashantgupta24:
> > I can manually add a prompt adapter model and check it in, that works well with the unit tests. Only thing is that it's for bloomz-560m, I couldn't find one for the default facebook-opt/125m
> 
> How large is that? We can add it as a tests which runs only with specific pytest flags (`--longtests` or something like that)

Honestly it's not too large, but it doesn't work well with cpu :( Runs great on the gpu dev pod

tests/conftest.py
prashantgupta24: I see
```
>>> vllm.config.is_cpu()
False
```

tests/test_grpc_server.py
```diff
@@ -24,15 +24,17 @@ def test_generation_request(grpc_client, grpc_server_thread_port):
...
+):
     streaming_response = grpc_client.make_request_stream(
         "The answer to life the universe and everything is ",
     )
 
     text_chunks: list[str] = [chunk.text for chunk in streaming_response]
 
     assert text_chunks
-    assert len(text_chunks) == 11
+    assert len(text_chunks) == 6 if prompt_available else 11
```
prashantgupta24, line 37: just because we load `bloom` for prompts

dtrifiro:
@prashantgupta24 there's a few conflicts, but CI should pass

prashantgupta24:
I think there's less reason for this to be merged now, since Nick is suggesting we encourage use of `adapter_config` type adapters instead of the `*.pt` ones.

dtrifiro:
Closing this then, feel free to reopen if anything changes, thanks!

PR #51, opened 2024-07-19
tests: fix lora tests handling

dtrifiro:
tests were being skipped because of the way fixtures were being used

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -39,24 +39,24 @@ def monkeysession():
 
 
 @pytest.fixture(scope="session")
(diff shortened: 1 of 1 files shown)


prashantgupta24 (approved):

dtrifiro:
Tests were failing because CI expects the cpu version to be installed, but https://github.com/opendatahub-io/vllm/pull/95 changed the odh vllm package name to `vllm-odh`, causing the upstream vllm to be installed.

This was reverted in https://github.com/opendatahub-io/vllm/pull/98.

PR #52, opened 2024-07-19
Add temporary validation to disallow seed with pipeline parallel

njhill:
Until we fix upstream bug

(Rebased on ~`fix-lora-toks`~ `fix-tokenize` to make the PRs stacked)

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -186,6 +186,9 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
         self.skip_special_tokens = not args.output_special_tokens
         self.default_include_stop_seqs = args.default_include_stop_seqs
 
+        # Temporary to validate parameters currently broken with PP
+        self.pipeline_parallel = args.pipeline_parallel_size > 1
+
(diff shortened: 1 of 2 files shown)


src/vllm_tgis_adapter/grpc/pb/generation.proto
```diff
@@ -225,6 +225,9 @@ message TokenInfo {
 
 message BatchedTokenizeRequest {
   string model_id = 1;
+  // Deprecated in favor of adapter_id
+  optional string prefix_id = 5;
```
dtrifiro, line 229: We cannot use `5` here as it's used by `truncate_input_tokens` below:

```
      vllm_tgis_adapter/grpc/pb/generation.proto:236:34: Field number 5 has already been used in "fmaas.BatchedTokenizeRequest" by field "prefix_id".
```

njhill: Thanks @dtrifiro my bad, pushed a fix now.

njhill:
@dtrifiro not a big deal but I was thinking we would have merged the fix-tokenize branch to main first, followed by merging this to main. Since they are unrelated changes. This was only targeted at the fix-tokenize branch so that we could have a single branch containing both to use in our immediate builds...

PR #59, opened 2024-07-23
feat: Add model-util CLI 

rafvasq:
- Supersedes https://github.com/IBM/vllm/pull/16, https://github.com/IBM/vllm/pull/52, https://github.com/opendatahub-io/vllm/pull/92

- Related to: https://github.com/vllm-project/vllm/pull/5090

Changes:
- Adds commands to the adapter using either entry point `model-util` or `text-generation-server` 
  * `model-util download-weights`
  * `model-util convert-to-safetensors`
  * `model-util convert-to-fast-tokenizer`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,7 @@ tests = [
 ]
 dev = [
   "vllm_tgis_adapter[tests]",
-  "ruff==0.5.4",
+  "ruff==0.5.5",
   "mypy==1.10.1",
   "types-protobuf==5.26.0.20240422",
   "types-requests==2.32.0.20240712"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ Source = "https://github.com/opendatahub-io/vllm_tgis_adapter"
 
 [project.scripts]
 grpc-healthcheck = "vllm_tgis_adapter.healthcheck:cli"
-vllm = "vllm_tgis_adapter.tgis_utils.scripts:cli"
+adapter = "vllm_tgis_adapter.tgis_utils.scripts:cli"
 
 [project.optional-dependencies]
 tests = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,6 +85,9 @@ vllm_tgis_adapter = [
 
 [tool.pytest.ini_options]
 addopts = "-ra"
+markers = [
+  "large"
+]
 
 [tool.coverage.run]
 branch = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -47,6 +47,7 @@ Source = "https://github.com/opendatahub-io/vllm-tgis-adapter"
 [project.scripts]
 grpc-healthcheck = "vllm_tgis_adapter.healthcheck:cli"
 model-util = "vllm_tgis_adapter.tgis_utils.scripts:cli"
+text-generation-server = "vllm_tgis_adapter.tgis_utils.scripts:cli"
 
 [project.optional-dependencies]
 tests = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,7 @@ Source = "https://github.com/opendatahub-io/vllm-tgis-adapter"
 
 [project.scripts]
 grpc-healthcheck = "vllm_tgis_adapter.healthcheck:cli"
-adapter = "vllm_tgis_adapter.tgis_utils.scripts:cli"
+model-util = "vllm_tgis_adapter.tgis_utils.scripts:cli"
 
 [project.optional-dependencies]
 tests = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -84,9 +84,9 @@ vllm_tgis_adapter = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-ra"
+addopts = "-ra -k \"not hf_data\""
 markers = [
-  "large"
+  "hf_data: marks tests that download data from HF hub (deselect with '-m \"not hf_data\"')"
 ]
 
 [tool.coverage.run]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,7 @@ Issues = "https://github.com/opendatahub-io/vllm-tgis-adapter/issues"
 Source = "https://github.com/opendatahub-io/vllm-tgis-adapter"
 
 [project.scripts]
-grpc-healthcheck = "vllm_tgis_adapter.healthcheck:cli"
+grpc_healthcheck = "vllm_tgis_adapter.healthcheck:cli"
 model-util = "vllm_tgis_adapter.tgis_utils.scripts:cli"
 text-generation-server = "vllm_tgis_adapter.tgis_utils.scripts:cli"
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,7 +45,8 @@ Issues = "https://github.com/opendatahub-io/vllm_tgis_adapter/issues"
 Source = "https://github.com/opendatahub-io/vllm_tgis_adapter"
 
 [project.scripts]
-grpc_healthcheck = "vllm_tgis_adapter.healthcheck:cli"
+grpc-healthcheck = "vllm_tgis_adapter.healthcheck:cli"
+vllm = "vllm_tgis_adapter.tgis_utils.scripts:cli"
 
 [project.optional-dependencies]
 tests = [
--- a/src/vllm_tgis_adapter/tgis_utils/hub.py
+++ b/src/vllm_tgis_adapter/tgis_utils/hub.py
@@ -1,0 +1,270 @@
+import concurrent
+import datetime
+import glob
+import json
+import logging
+import os
+from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+from pathlib import Path
+from typing import Dict, List, Optional
+
+import torch
+from huggingface_hub import HfApi, hf_hub_download, try_to_load_from_cache
+from huggingface_hub.utils import LocalEntryNotFoundError
+from safetensors.torch import (_find_shared_tensors, _is_complete, load_file,
+                               save_file)
+from tqdm import tqdm
+
+TRUST_REMOTE_CODE = os.getenv("TRUST_REMOTE_CODE") == "true"
+logger = logging.getLogger(__name__)
+
+
+def weight_hub_files(model_name,
+                     extension=".safetensors",
+                     revision=None,
+                     auth_token=None):
+    """Get the safetensors filenames on the hub"""
+    exts = [extension] if isinstance(extension, str) else extension
+    api = HfApi()
+    info = api.model_info(model_name, revision=revision, token=auth_token)
+    filenames = [
+        s.rfilename for s in info.siblings if any(
+            s.rfilename.endswith(ext) and len(s.rfilename.split("/")) == 1
+            and "arguments" not in s.rfilename and "args" not in s.rfilename
+            and "training" not in s.rfilename for ext in exts)
+    ]
+    return filenames
+
+
+def weight_files(model_name, extension=".safetensors", revision=None):
+    """Get the local safetensors filenames"""
+    filenames = weight_hub_files(model_name, extension)
+    files = []
+    for filename in filenames:
+        cache_file = try_to_load_from_cache(model_name,
+                                            filename=filename,
+                                            revision=revision)
+        if cache_file is None:
+            raise LocalEntryNotFoundError(
+                f"File {filename} of model {model_name} not found in "
+                f"{os.getenv('HUGGINGFACE_HUB_CACHE', 'the local cache')}. "
+                f"Please run `vllm \
+                    download-weights {model_name}` first.")
+        files.append(cache_file)
+
+    return files
+
+
+def download_weights(model_name,
+                     extension=".safetensors",
+                     revision=None,
+                     auth_token=None):
+    """Download the safetensors files from the hub"""
+    filenames = weight_hub_files(model_name,
+                                 extension,
+                                 revision=revision,
+                                 auth_token=auth_token)
+
+    download_function = partial(
+        hf_hub_download,
+        repo_id=model_name,
+        local_files_only=False,
+        revision=revision,
+        token=auth_token,
+    )
+
+    print(f"Downloading {len(filenames)} files for model {model_name}")
+    executor = ThreadPoolExecutor(max_workers=5)
+    futures = [
+        executor.submit(download_function, filename=filename)
+        for filename in filenames
+    ]
+    files = [
+        future.result()
+        for future in tqdm(concurrent.futures.as_completed(futures),
+                           total=len(futures))
+    ]
+
+    return files
+
+
+def get_model_path(model_name: str, revision: Optional[str] = None):
+    """Get path to model dir in local huggingface hub (model) cache"""
+    config_file = "config.json"
+    err = None
+    try:
+        config_path = try_to_load_from_cache(
+            model_name,
+            config_file,
+            cache_dir=os.getenv("TRANSFORMERS_CACHE"
+                                ),  # will fall back to HUGGINGFACE_HUB_CACHE
+            revision=revision,
+        )
+        if config_path is not None:
+            return config_path.removesuffix(f"/{config_file}")
+    except ValueError as e:
+        err = e
+
+    if os.path.isfile(f"{model_name}/{config_file}"):
+        return model_name  # Just treat the model name as an explicit model path
+
+    if err is not None:
+        raise err
+
+    raise ValueError(
+        f"Weights not found in local cache for model {model_name}")
+
+
+def local_weight_files(model_path: str, extension=".safetensors"):
+    """Get the local safetensors filenames"""
+    ext = "" if extension is None else extension
+    return glob.glob(f"{model_path}/*{ext}")
+
+
+def local_index_files(model_path: str, extension=".safetensors"):
+    """Get the local .index.json filename"""
+    ext = "" if extension is None else extension
+    return glob.glob(f"{model_path}/*{ext}.index.json")
+
+
+def _remove_duplicate_names(
+    state_dict: Dict[str, torch.Tensor],
+    *,
+    preferred_names: List[str] = None,
+    discard_names: List[str] = None,
+) -> Dict[str, List[str]]:
+    if preferred_names is None:
+        preferred_names = []
+    preferred_names = set(preferred_names)
+    if discard_names is None:
+        discard_names = []
+    discard_names = set(discard_names)
+
+    shareds = _find_shared_tensors(state_dict)
+    to_remove = defaultdict(list)
+    for shared in shareds:
+        # _find_shared_tensors returns a list of sets of names of tensors that
+        # have the same data, including sets with one element that aren't shared
+        if len(shared) == 1:
+            continue
+
+        complete_names = set(
+            [name for name in shared if _is_complete(state_dict[name])])
+        if not complete_names:
+            raise RuntimeError(f"Error while trying to find names to remove \
+                    to save state dict, but found no suitable name to \
+                        keep for saving amongst: {shared}. None is covering \
+                            the entire storage.Refusing to save/load the model \
+                                since you could be storing much more \
+                                    memory than needed. Please refer to\
+            https://huggingface.co/docs/safetensors/torch_shared_tensors \
+                                            for more information. \
+                                                Or open an issue.")
+
+        keep_name = sorted(list(complete_names))[0]
+
+        # Mechanism to preferentially select keys to keep
+        # coming from the on-disk file to allow
+        # loading models saved with a different choice
+        # of keep_name
+        preferred = complete_names.difference(discard_names)
+        if preferred:
+            keep_name = sorted(list(preferred))[0]
+
+        if preferred_names:
+            preferred = preferred_names.intersection(complete_names)
+            if preferred:
+                keep_name = sorted(list(preferred))[0]
+        for name in sorted(shared):
+            if name != keep_name:
+                to_remove[keep_name].append(name)
+    return to_remove
+
+
+def convert_file(pt_file: Path, sf_file: Path, discard_names: List[str]):
+    """
+    Convert a pytorch file to a safetensors file
+    This will remove duplicate tensors from the file.
+
+    Unfortunately, this might not respect *transformers* convention.
+    Forcing us to check for potentially different keys during load when looking
+    for specific tensors (making tensor sharing explicit).
+    """
+    loaded = torch.load(pt_file, map_location="cpu")
+    if "state_dict" in loaded:
+        loaded = loaded["state_dict"]
+    to_removes = _remove_duplicate_names(loaded, discard_names=discard_names)
+
+    metadata = {"format": "pt"}
+    for kept_name, to_remove_group in to_removes.items():
+        for to_remove in to_remove_group:
+            if to_remove not in metadata:
+                metadata[to_remove] = kept_name
+            del loaded[to_remove]
+    # Force tensors to be contiguous
+    loaded = {k: v.contiguous() for k, v in loaded.items()}
+
+    dirname = os.path.dirname(sf_file)
+    os.makedirs(dirname, exist_ok=True)
+    save_file(loaded, sf_file, metadata=metadata)
+    reloaded = load_file(sf_file)
+    for k in loaded:
+        pt_tensor = loaded[k]
+        sf_tensor = reloaded[k]
+        if not torch.equal(pt_tensor, sf_tensor):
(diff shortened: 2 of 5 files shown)


pyproject.toml
```diff
@@ -45,7 +45,8 @@ Issues = "https://github.com/opendatahub-io/vllm_tgis_adapter/issues"
 Source = "https://github.com/opendatahub-io/vllm_tgis_adapter"
 
 [project.scripts]
-grpc_healthcheck = "vllm_tgis_adapter.healthcheck:cli"
+grpc-healthcheck = "vllm_tgis_adapter.healthcheck:cli"
+vllm = "vllm_tgis_adapter.tgis_utils.scripts:cli"
```
prashantgupta24, line 49: won't this over-write the original `vllm` CLI?

rafvasq: I was wondering, should I use something like `tgis` instead since we're treating these as their own?

prashantgupta24: Looking at this review comment from Nick, Ideally they should all be part of `vllm` itself
https://github.com/IBM/vllm/pull/16#pullrequestreview-1993895084. 

If we can't think of a way to override the name, I'm not sure if `tgis` makes sense though, might confuse people? The commands have nothing to do with `tgis`, they're just ported over from `tgis` which wouldn't make sense 🤔

prashantgupta24: Also we should totally add a subcommand that will start the server using this adapter

prashantgupta24: Right now having to type `python3 -m vllm_tgis_adapter` is a pain

prashantgupta24:
Also how did you test this? Can we do something like what Nick suggested? 
```
Have you tried building an image with this and running the tools in there?
```

pyproject.toml
prashantgupta24, line 1: Also how did you test this? Can we do something like what Nick suggested? 
```
Have you tried building an image with this and running the tools in there?
```

pyproject.toml
dtrifiro: We can add a `vllm_tgis_adapter` cli entrypoint if required. If you want to save a character you can skip the `3` in `python3` though 😉

prashantgupta24: Yeah if we're not able to extend the original `vllm` cli then an adapter cli would be great

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,274 @@
...
+from pathlib import Path
+from typing import Dict, List, Optional
+
+import torch
+from huggingface_hub import HfApi, hf_hub_download, try_to_load_from_cache
+from huggingface_hub.utils import LocalEntryNotFoundError
+from safetensors.torch import _find_shared_tensors, _is_complete, load_file, save_file
+from tqdm import tqdm
+
+TRUST_REMOTE_CODE = os.getenv("TRUST_REMOTE_CODE") == "true"
```
dtrifiro, line 19: This looks unused?

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,274 @@
...
+    to_remove = defaultdict(list)
+    for shared in shareds:
+        # _find_shared_tensors returns a list of sets of names of tensors that
+        # have the same data, including sets with one element that aren't shared
+        if len(shared) == 1:
+            continue
+
+        complete_names = {name for name in shared if _is_complete(state_dict[name])}
+        if not complete_names:
+            raise RuntimeError(f"Error while trying to find names to remove \
```
dtrifiro, line 156: This will result in a pretty weirdly formatted error message, please format it like so:

```python
    raise Exception(
        "this is a pretty long"
        " error message"
        " which is more than 88 columns long"
        " so it needs to be split along multiple lines"
    )

```

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,274 @@
...
+import datetime
+import glob
+import json
+import logging
+import os
+from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
+from functools import partial
+from pathlib import Path
+from typing import Dict, List, Optional
```
dtrifiro, line 11: We should be able to use native types if `from __future__ import annotations` at the top of this file


and replacing `Dict` -> `dict`, `List` -> `list` and `Optional[<type>]` -> `<type> | None`

dtrifiro:
Please see the [contributing guide](https://github.com/opendatahub-io/vllm-tgis-adapter/blob/main/README.md#contributing) for information on how to set up `pre-commit` so that CI does not fail

pyproject.toml
rafvasq: I tested by using `pip install git+https://github.com/rafvasq/vllm-tgis-adapter@add-cli-cmds`. I haven't had much luck building the `Dockerfile.ubi` image (on a mac),  but here are some logs in its current state:

**download-weights**
```
$ adapter download-weights --extension ".bin" EleutherAI/gpt-neo-125m
INFO 07-31 16:39:09 scripts.py:46] .bin
INFO 07-31 16:39:09 hub.py:88] Downloading 8 files for model EleutherAI/gpt-neo-125m
generation_config.json: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 119/119 [00:00<00:00, 1.25MB/s]
special_tokens_map.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 357/357 [00:00<00:00, 3.51MB/s]
README.md: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4.10k/4.10k [00:00<00:00, 34.2MB/s]
config.json: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1.01k/1.01k [00:00<00:00, 13.0MB/s]
vocab.json: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 899k/899k [00:00<00:00, 54.0MB/s]
tokenizer_config.json: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 727/727 [00:00<00:00, 9.59MB/s]
tokenizer.json: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2.11M/2.11M [00:00<00:00, 65.7MB/s]
pytorch_model.bin: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 526M/526M [00:01<00:00, 361MB/s]
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 8/8 [00:01<00:00,  5.21it/s]
INFO 07-31 16:39:11 scripts.py:168] Model EleutherAI/gpt-neo-125m already has a fast tokenizer
```

**convert-to-safetensors**
```
$ adapter convert-to-safetensors EleutherAI/gpt-neo-125m
INFO 07-31 16:55:35 hub.py:217] Converting 1 pytorch .bin files to .safetensors...
INFO 07-31 16:55:35 hub.py:221] Converting: [1] "pytorch_model.bin"
INFO 07-31 16:55:36 hub.py:225] Converted: [1] "model.safetensors" -- Took: 0 seconds
```

```
$ rm models--HuggingFaceH4--zephyr-7b-alpha/snapshots/2ce2d025864af849b3e5029e2ec9d568eeda892d/model.safetensors.index.json
$ adapter convert-to-safetensors HuggingFaceH4/zephyr-7b-alpha
INFO 07-31 17:00:56 hub.py:177] Converting pytorch .bin.index.json files to .safetensors.index.json
INFO 07-31 17:00:56 hub.py:217] Converting 8 pytorch .bin files to .safetensors...
INFO 07-31 17:00:56 hub.py:221] Converting: [0] "pytorch_model-00002-of-00008.bin"
INFO 07-31 17:00:58 hub.py:225] Converted: [0] "model-00002-of-00008.safetensors" -- Took: 1 seconds
INFO 07-31 17:00:58 hub.py:221] Converting: [0] "pytorch_model-00004-of-00008.bin"
INFO 07-31 17:01:00 hub.py:225] Converted: [0] "model-00004-of-00008.safetensors" -- Took: 1 seconds
INFO 07-31 17:01:00 hub.py:221] Converting: [0] "pytorch_model-00003-of-00008.bin"
INFO 07-31 17:01:02 hub.py:225] Converted: [0] "model-00003-of-00008.safetensors" -- Took: 1 seconds
INFO 07-31 17:01:02 hub.py:221] Converting: [0] "pytorch_model-00001-of-00008.bin"
INFO 07-31 17:01:04 hub.py:225] Converted: [0] "model-00001-of-00008.safetensors" -- Took: 1 seconds
INFO 07-31 17:01:04 hub.py:221] Converting: [0] "pytorch_model-00005-of-00008.bin"
INFO 07-31 17:01:06 hub.py:225] Converted: [0] "model-00005-of-00008.safetensors" -- Took: 1 seconds
INFO 07-31 17:01:06 hub.py:221] Converting: [0] "pytorch_model-00006-of-00008.bin"
INFO 07-31 17:01:07 hub.py:225] Converted: [0] "model-00006-of-00008.safetensors" -- Took: 1 seconds
INFO 07-31 17:01:07 hub.py:221] Converting: [0] "pytorch_model-00008-of-00008.bin"
INFO 07-31 17:01:08 hub.py:225] Converted: [0] "model-00008-of-00008.safetensors" -- Took: 0 seconds
INFO 07-31 17:01:08 hub.py:221] Converting: [1] "pytorch_model-00007-of-00008.bin"
INFO 07-31 17:01:10 hub.py:225] Converted: [1] "model-00007-of-00008.safetensors" -- Took: 1 seconds
```

**convert-to-fast-tokenizer**

```
$ rm models--EleutherAI--gpt-neo-125m/snapshots/21def0189f5705e2521767faed922f1f15e7d7db/tokenizer.json
$ adapter convert-to-fast-tokenizer  EleutherAI/gpt-neo-125m
merges.txt: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 456k/456k [00:00<00:00, 477kB/s]
INFO 07-31 17:01:32 scripts.py:185] Saved tokenizer to /tmp/cache/models--EleutherAI--gpt-neo-125m/snapshots/21def0189f5705e2521767faed922f1f15e7d7db
```

dtrifiro:
vllm@main CI will keep failing until #68 solves the problem and is merged

prashantgupta24 (approved):

prashantgupta24:
@rafvasq can you trim down the commits?

rafvasq:
Trimmed them down, @prashantgupta24

njhill:
Thanks @rafvasq, and thanks @prashantgupta24 @dtrifiro for all of the reviews. This is looking great!

I know there was already some discussion about this above but I'm wondering whether we should make the command something different. Maybe even just make it `text-generation-server` since these are mainly for backwards compatibility?

Or even something like `model-util`? Even better would I guess would be to have both, i.e. keep `text-generation-server` as an alias if that's possible, for backwards compatibility, but we can just document the new command.

prashantgupta24:
> Thanks @rafvasq, and thanks @prashantgupta24 @dtrifiro for all of the reviews. This is looking great!
> 
> I know there was already some discussion about this above but I'm wondering whether we should make the command something different. Maybe even just make it `text-generation-server` since these are mainly for backwards compatibility?
> 
> Or even something like `model-util`? Even better would I guess would be to have both, i.e. keep `text-generation-server` as an alias if that's possible, for backwards compatibility, but we can just document the new command.

I was a bit hesitant in using `text-generation-server` just because it might confuse folks who have worked with TGIS earlier?

njhill:
> > Thanks @rafvasq, and thanks @prashantgupta24 @dtrifiro for all of the reviews. This is looking great!
> > I know there was already some discussion about this above but I'm wondering whether we should make the command something different. Maybe even just make it `text-generation-server` since these are mainly for backwards compatibility?
> > Or even something like `model-util`? Even better would I guess would be to have both, i.e. keep `text-generation-server` as an alias if that's possible, for backwards compatibility, but we can just document the new command.
> 
> I was a bit hesitant in using `text-generation-server` just because it might confuse folks who have worked with TGIS earlier?

Imho it shouldn't introduce confusion, just would hopefully allow their existing scripts/processes to work without change. But if it's not straightforward to make an alias then I guess not a huge deal.

dtrifiro:
@rafvasq There's a few conflicts, could you rebase this on main? Thanks!

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,230 @@
...
+    download_function = partial(
+        hf_hub_download,
+        repo_id=model_name,
+        local_files_only=False,
+        revision=revision,
+        token=auth_token,
+    )
+
+    logger.info("Downloading %s files for model %s", len(filenames), model_name)
+    executor = ThreadPoolExecutor(max_workers=5)
```
dtrifiro, line 89: Why 5?

```suggestion
    executor = ThreadPoolExecutor(max_workers=os.cpu_count())
```

dtrifiro: Even though this could wreak havoc for machines with a large amount of threads, how about:

```python
max_workers=min(16, os.cpu_count())
```

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,230 @@
...
+    if Path(f"{model_name}/{config_file}").is_file():
+        return model_name  # Just treat the model name as an explicit model path
+
+    if err is not None:
+        raise err
+
+    raise ValueError(f"Weights not found in local cache for model {model_name}")
+
+
+def local_weight_files(model_path: str, extension: str = ".safetensors") -> list:
```
dtrifiro, line 128:
```suggestion
def local_weight_files(model_path: str, extension: str = ".safetensors") -> list[Path]:
```

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,230 @@
...
+    """Get the local safetensors filenames."""
+    ext = "" if extension is None else extension
+    return list(Path(f"{model_path}").glob(f"*{ext}"))
+
+
+def local_index_files(model_path: str, extension: str = ".safetensors") -> list:
```
dtrifiro, line 134:
```suggestion
def local_index_files(model_path: str, extension: str = ".safetensors") -> list[Path]
```

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,230 @@
...
+    for kept_name, to_remove_group in to_removes.items():
+        for to_remove in to_remove_group:
+            if to_remove not in metadata:
+                metadata[to_remove] = kept_name
+            del loaded[to_remove]
+    # Force tensors to be contiguous
+    loaded = {k: v.contiguous() for k, v in loaded.items()}
+
+    dirname = Path(sf_file).parent
+    Path(dirname).mkdir(parents=True, exist_ok=True)
```
dtrifiro, line 162:
```suggestion
    sf_file.parent.mkdir(parents=True, exist_ok=True)
```

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,230 @@
...
+    logger.info("Converting pytorch .bin.index.json files to .safetensors.index.json")
+    with open(source_file) as f:
+        index = json.load(f)
+
+    index["weight_map"] = {
+        k: weight_file_map[v] for k, v in index["weight_map"].items()
+    }
+
+    with open(dest_file, "w") as f:
+        json.dump(index, f, indent=4)
```
dtrifiro, line 186: Do we care about human readability here? If not:

```suggestion
        json.dump(index, f)
```

src/vllm_tgis_adapter/tgis_utils/scripts.py
```diff
@@ -0,0 +1,238 @@
+# The CLI entrypoint to vLLM.
```
dtrifiro, line 1:
```suggestion
```

src/vllm_tgis_adapter/tgis_utils/hub.py
```diff
@@ -0,0 +1,230 @@
...
+        if config_path is not None:
+            return config_path.removesuffix(f"/{config_file}")
+    except ValueError as e:
+        err = e
+
+    if Path(f"{model_name}/{config_file}").is_file():
+        return model_name  # Just treat the model name as an explicit model path
+
+    if err is not None:
+        raise err
```
dtrifiro, line 123: Why do we need this `try/except` setting `err=` to raise it below?

```suggestion
    config_path = try_to_load_from_cache(
        model_name,
        config_file,
        cache_dir=os.getenv(
            "TRANSFORMERS_CACHE"
        ),  # will fall back to HUGGINGFACE_HUB_CACHE
        revision=revision,
    )
    if config_path is not None:
        return config_path.removesuffix(f"/{config_file}")
    if Path(f"{model_name}/{config_file}").is_file():
        return model_name  # Just treat the model name as an explicit model pat
```

src/vllm_tgis_adapter/tgis_utils/scripts.py
```diff
@@ -0,0 +1,237 @@
+from __future__ import annotations
+
+from pathlib import Path
+from typing import TYPE_CHECKING
+
+from vllm.scripts import registrer_signal_handlers
```
dtrifiro, line 6: Do we actually need these signal handlers?

dtrifiro (approved):

pyproject.toml
```diff
@@ -45,7 +45,9 @@ Issues = "https://github.com/opendatahub-io/vllm-tgis-adapter/issues"
 Source = "https://github.com/opendatahub-io/vllm-tgis-adapter"
 
 [project.scripts]
-grpc_healthcheck = "vllm_tgis_adapter.healthcheck:cli"
+grpc-healthcheck = "vllm_tgis_adapter.healthcheck:cli"
```
dtrifiro, line 48: Please do not change this:
```suggestion
grpc_healthcheck = "vllm_tgis_adapter.healthcheck:cli"
```

dtrifiro:
@rafvasq Please do not mark conversations as resolved without either addressing them or replying, thanks!

src/vllm_tgis_adapter/tgis_utils/hub.py
rafvasq: Updated

src/vllm_tgis_adapter/tgis_utils/hub.py
rafvasq: Updated

src/vllm_tgis_adapter/tgis_utils/hub.py
rafvasq: Updated

src/vllm_tgis_adapter/tgis_utils/hub.py
rafvasq: Updated, don't believe we care about human readability here

src/vllm_tgis_adapter/tgis_utils/scripts.py
rafvasq: Updated

src/vllm_tgis_adapter/tgis_utils/hub.py
rafvasq: Updated to your suggestion

src/vllm_tgis_adapter/tgis_utils/scripts.py
rafvasq: Don't believe so, removed them

njhill:
> > > Thanks @rafvasq, and thanks @prashantgupta24 @dtrifiro for all of the reviews. This is looking great!
> > > I know there was already some discussion about this above but I'm wondering whether we should make the command something different. Maybe even just make it `text-generation-server` since these are mainly for backwards compatibility?
> > > Or even something like `model-util`? Even better would I guess would be to have both, i.e. keep `text-generation-server` as an alias if that's possible, for backwards compatibility, but we can just document the new command.
> > 
> > 
> > I was a bit hesitant in using `text-generation-server` just because it might confuse folks who have worked with TGIS earlier?
> 
> Imho it shouldn't introduce confusion, just would hopefully allow their existing scripts/processes to work without change. But if it's not straightforward to make an alias then I guess not a huge deal.

@rafvasq @prashantgupta24 thinking some more about this, we should add the `text-generation-server` alias if possible. Would like this to be a drop-in for the existing configs/automation that the software team have.

rafvasq:
@dtrifiro oh, if by addressing them you mean making the requested changes then I had done that, which is why I resolved them! I left them open now with an explicit comment though.

rafvasq:
@njhill, sorry I forgot to mention it but I did introduce the `text-generation-server` alias:

https://github.com/opendatahub-io/vllm-tgis-adapter/pull/59/files#diff-50c86b7ed8ac2cf95bd48334961bf0530cdc77b5a56f852c5c61b89d735fd711R50

tests/test_hub.py
```diff
@@ -0,0 +1,52 @@
...
+
+def test_weight_hub_files_empty():
+    filenames = weight_hub_files("bigscience/bloom", ".errors")
+    assert filenames == []
+
+
+def test_download_weights():
+    files = download_weights("bigscience/bloom-560m")
+    local_files = weight_files("bigscience/bloom-560m")
+    assert files == local_files
```
dtrifiro, line 47: This test tries to download 350GB worth of files, That's way too much to run in a CI pipeline, I'd prefer to either use something way smaller, or make all tests which download lots of data from hf hub hidden behind a custom marker:

https://docs.pytest.org/en/latest/example/markers.html

src/vllm_tgis_adapter/tgis_utils/scripts.py
```diff
@@ -0,0 +1,234 @@
...
+    if Path.exists(Path(model_path) / "tokenizer.json"):
+        logger.info("Model %s already has a fast tokenizer", model_name)
+        return
+
+    if output_path is not None:
+        if not Path.isdir(output_path):
+            logger.info("Output path %s must exist and be a directory", output_path)
+            return
+    else:
+        output_path = model_path
```
dtrifiro, line 172: Nit: 
```suggestion
    if output_path is not None and not Path.isdir(output_path):
        logger.info("Output path %s must exist and be a directory", output_path)
        return
    output_path = model_path
```

dtrifiro:
Thanks @rafvasq. This looks very good at this point, just a few nits left

tests/test_hub.py
prashantgupta24: Should be possible to use a tiny model to test out this scenario

src/vllm_tgis_adapter/tgis_utils/scripts.py
rafvasq: Fixed this

tests/test_hub.py
prashantgupta24: It would be good to have unit tests for the functions that emulate the behavior somehow (can be tackled in subsequent PRs) - worth looking into the custom marker

rafvasq: Went ahead and added a custom marker `large` to this test, although it seems a bit generic. If there's a preferred name just let me know!

tests/test_hub.py
```diff
@@ -0,0 +1,53 @@
...
+        for p in local_pt_files
+    ]
+    convert_files(local_pt_files, local_st_files, discard_names=[])
+
+    found_st_files = weight_files(model_id)
+
+    assert all(str(p) in found_st_files for p in local_st_files)
+
+
+def test_weight_hub_files():
```
dtrifiro, line 29: Marks should be added to other tests aswell.

pyproject.toml
```diff
@@ -83,6 +85,9 @@ vllm_tgis_adapter = [
 
 [tool.pytest.ini_options]
 addopts = "-ra"
+markers = [
```
dtrifiro, line 88: Will also need to configure pytest to skip tests marked with large:

https://docs.pytest.org/en/7.1.x/example/markers.html#custom-marker-and-command-line-option-to-control-test-runs

rafvasq: I've updated the marker to `hf_data` to capture under the assumption that it's not just about avoiding large model downloads but all HF data which can add up -- configured the test to exclude these

tests/test_hub.py
rafvasq: Since all of the tests in the file call for HF data I used the marker global variable on the module:

https://docs.pytest.org/en/7.1.x/reference/reference.html#globalvar-pytestmark

dtrifiro: Good call

dtrifiro (approved):

PR #62, opened 2024-07-24
deps: remove transformers

dtrifiro:
This is an upstream dependency


dtrifiro:
Done in https://github.com/opendatahub-io/vllm-tgis-adapter/pull/60

PR #63, opened 2024-07-24
:goal_net: Kill servers on engine death

joerunde:
## Description
See https://github.com/vllm-project/vllm/pull/6594

This PR modifies the grpc server's exception handler to check if the async llm engine is still running.
If it's not then the handler sends a SIGTERM to kill the serving process, which should trigger the graceful shutdown handlers.

This approach is a bit heavy-handed, but `await server.stop(0)` seems very unhappy if called within the context of a request handler. Maybe we could use some grpc context termination callbacks for this, but I didn't really think it was worth the time to investigate that too deeply. I think the only real drawback here is that unit testing this feature would be pretty difficult with an `os.kill` in the mix.

This PR currently does _not_ add this handling to the http server, but once PRs 6594 and https://github.com/vllm-project/vllm/pull/6740 are merged, we should be able to remove our copy-pasta http server so both will handle engine death properly.

## How Has This Been Tested?
Tested by installing vllm==0.5.3.post1 and the adapter from this branch onto a dev pod, injecting a runtime failure into the llm engine, and running a grpc request.

## Merge criteria:

### DIFF
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
-import signal
 from concurrent.futures import FIRST_COMPLETED
 from typing import TYPE_CHECKING
 
@@ -35,6 +34,7 @@ async def start_servers(args: argparse.Namespace) -> None:
             run_http_server(args, engine),
             name="http_server",
         )
+        # The http server task will catch interrupt signals for us
         tasks.append(http_server_task)
 
         grpc_server_task = loop.create_task(
@@ -43,28 +43,16 @@ async def start_servers(args: argparse.Namespace) -> None:
         )
         tasks.append(grpc_server_task)
 
-        def signal_handler() -> None:
-            # prevents the uvicorn signal handler to exit early
-            for task in tasks:
-                task.cancel()
-
-        async def override_signal_handler() -> None:
-            loop = asyncio.get_running_loop()
-
-            for sig in (signal.SIGINT, signal.SIGTERM):
-                loop.add_signal_handler(sig, signal_handler)
-
-        await override_signal_handler()
-
         with contextlib.suppress(asyncio.CancelledError):
             # Both server tasks will exit normally on shutdown, so we await
             # FIRST_COMPLETED to catch either one shutting down.
-            await asyncio.wait(
-                tasks,
-                return_when=FIRST_COMPLETED,
-            )
+            await asyncio.wait(tasks, return_when=FIRST_COMPLETED)
+        # Once either server shuts down, cancel the other
         for task in tasks:
             task.cancel()
+
+        # Final wait for both servers to finish
+        await asyncio.wait(tasks)
 
         check_for_failed_tasks(tasks)
 
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -3,7 +3,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import signal
-from concurrent.futures import FIRST_EXCEPTION
+from concurrent.futures import FIRST_COMPLETED
 from typing import TYPE_CHECKING
 
 import uvloop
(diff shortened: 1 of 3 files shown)


dtrifiro:
Waiting on https://github.com/vllm-project/vllm/pull/6594 to make sure that we have a common way of stopping the server

njhill:
https://github.com/vllm-project/vllm/pull/6594 now merged finally

joerunde:
https://github.com/vllm-project/vllm/pull/6594 is now merged- I'll see if I can take another pass at this to not use os.kill

src/vllm_tgis_adapter/__main__.py
```diff
@@ -54,14 +54,15 @@ async def override_signal_handler() -> None:
             for sig in (signal.SIGINT, signal.SIGTERM):
                 loop.add_signal_handler(sig, signal_handler)
 
-        tasks.append(loop.create_task(override_signal_handler()))
+        await override_signal_handler()
 
         with contextlib.suppress(asyncio.CancelledError):
+            # Both server tasks will exit normally on shutdown, so we await
+            # FIRST_COMPLETED to catch either one shutting down.
```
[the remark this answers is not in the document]
joerunde: The engine dead error is caught by each individually, so when one stops we have to kill the other from here by cancelling the task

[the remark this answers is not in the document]
njhill: OK, just wondering about the normal/clean shutdown case

src/vllm_tgis_adapter/__main__.py
```diff
@@ -54,14 +54,15 @@ async def override_signal_handler() -> None:
             for sig in (signal.SIGINT, signal.SIGTERM):
                 loop.add_signal_handler(sig, signal_handler)
 
-        tasks.append(loop.create_task(override_signal_handler()))
+        await override_signal_handler()
```
dtrifiro, line 57: Given the way that signal handlers are set in `serve_http` [here](https://github.com/opendatahub-io/vllm/blob/main/vllm/entrypoints/launcher.py#L29-L42), I think this will be overridden.

This is not an issue though: we can remove both `signal_handler()` and `override_signal_handler()`.

In this case ctrl-C will be handled by the http server, stopping the server task on ctrl-c. Then `asyncio.wait(tasks, return_when=FIRST_COMPLETED)` will return and both tasks will be cancelled in the `for task in tasks` loop below.

Here's the patch:
```diff
diff --git a/src/vllm_tgis_adapter/__main__.py b/src/vllm_tgis_adapter/__main__.py
index 8021dde..139f91b 100644
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -43,26 +43,9 @@ async def start_servers(args: argparse.Namespace) -> None:
         )
         tasks.append(grpc_server_task)
 
-        def signal_handler() -> None:
-            # prevents the uvicorn signal handler to exit early
-            for task in tasks:
-                task.cancel()
-
-        async def override_signal_handler() -> None:
-            loop = asyncio.get_running_loop()
-
-            for sig in (signal.SIGINT, signal.SIGTERM):
-                loop.add_signal_handler(sig, signal_handler)
-
-        await override_signal_handler()
-
         with contextlib.suppress(asyncio.CancelledError):
-            # Both server tasks will exit normally on shutdown, so we await
-            # FIRST_COMPLETED to catch either one shutting down.
-            await asyncio.wait(
-                tasks,
-                return_when=FIRST_COMPLETED,
-            )
+            await asyncio.wait(tasks, return_when=FIRST_COMPLETED)
+
         for task in tasks:
             task.cancel()
 ```

File: src/vllm_tgis_adapter/http.py [hunk 3]
```diff
@@ -57,6 +31,7 @@ async def run_http_server(
 
     await serve_http(
         app,
+        engine,
```
dtrifiro, line 34: This has been added after 0.5.4, so we need a backwards compatible way to handle this until a new version is released

src/vllm_tgis_adapter/__main__.py
joerunde: oh yeah, you're right. More negative diff!

src/vllm_tgis_adapter/http.py
joerunde: ah F :(

src/vllm_tgis_adapter/__main__.py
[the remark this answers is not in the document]
joerunde: ah, yeah this bit is just to catch either shutdown, we probably want a final await after this to make sure both stop cleanly

njhill (approved):
Thanks @joerunde!

PR #66, opened 2024-07-29
Make ADD_SPECIAL_TOKENS true by default

maxdebayser:
Before this change the ADD_SPECIAL_TOKENS acted as a tristate variable where the default (not set) would be to add special tokens only if the model didn't have a chat template.

However, this this default broke existing integration tests so, after some deliberation,  the decision was made to make the default to be "true". Due to the way that environment variables work, this means that this setting now only has 2 states: true and false.

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -85,9 +85,10 @@ _F = TypeVar("_F", Callable, Coroutine)
 logger = init_logger(__name__)
 service_metrics = ServiceMetrics()
 
-ADD_SPECIAL_TOKENS: str | None = os.getenv("ADD_SPECIAL_TOKENS")
-if ADD_SPECIAL_TOKENS is not None:
-    ADD_SPECIAL_TOKENS = ADD_SPECIAL_TOKENS.lower() not in (0, "false")
+ADD_SPECIAL_TOKENS: bool = os.getenv("ADD_SPECIAL_TOKENS", "true").lower() not in (
+    "0",
+    "false",
+)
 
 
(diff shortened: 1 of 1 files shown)


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -95,9 +95,7 @@
 logger = init_logger(__name__)
 service_metrics = ServiceMetrics()
 
-ADD_SPECIAL_TOKENS = os.getenv("ADD_SPECIAL_TOKENS")
-if ADD_SPECIAL_TOKENS is not None:
-    ADD_SPECIAL_TOKENS = ADD_SPECIAL_TOKENS.lower() not in (0, "false")
+ADD_SPECIAL_TOKENS = os.getenv("ADD_SPECIAL_TOKENS", "true").lower() not in (0, "false")
```
dtrifiro, line 98: Since `os.getenv` always returns a string, I think this should be 
```suggestion
ADD_SPECIAL_TOKENS = os.getenv("ADD_SPECIAL_TOKENS", "true").lower() not in ("0", "false")
```

prashantgupta24 (approved):
LGTM!

src/vllm_tgis_adapter/grpc/grpc_server.py
prashantgupta24: Good point!

njhill: Thanks, yes this was my mistake originally!

njhill (approved):
Thanks @maxdebayser

dtrifiro (approved):

PR #82, opened 2024-08-06
updates for vLLM==0.5.4

dtrifiro:
adaptations for vLLM==0.5.4

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -82,7 +82,7 @@ vllm_tgis_adapter = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-ra -W error::grpc.experimental.ExperimentalApiWarning"
+addopts = "-ra"
 
 [tool.coverage.run]
 branch = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 requires-python = ">=3.9"
 dynamic = ["version"]
 dependencies = [
-  "vllm>=0.5.3.post1",
+  "vllm>=0.5.4",
   "prometheus_client==0.20.0",
   "grpcio==1.62.2",
   "grpcio-health-checking==1.62.2",
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import signal
+from concurrent.futures import FIRST_EXCEPTION
 from typing import TYPE_CHECKING
 
 import uvloop
@@ -53,11 +54,23 @@ async def start_servers(args: argparse.Namespace) -> None:
 
         tasks.append(loop.create_task(override_signal_handler()))
 
-        try:
-            await asyncio.wait(tasks)
-        except asyncio.CancelledError:
-            for task in tasks:
-                task.cancel()
+        done, pending = await asyncio.wait(
+            tasks,
+            return_when=FIRST_EXCEPTION,
+        )
+        for task in pending:
+            task.cancel()
+
+        while done:
+            task = done.pop()
+            exc = task.exception()
+            if not exc:
+                continue
+
+            name = task.get_name()
+            coro_name = task.get_coro().__name__
+
+            raise RuntimeError(f"task={name} ({coro_name})") from exc
 
 
 if __name__ == "__main__":
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -1,313 +1,67 @@
 from __future__ import annotations
 
 import asyncio
-import importlib
-import inspect
-import re
 import signal
-from contextlib import asynccontextmanager
-from http import HTTPStatus
 from typing import TYPE_CHECKING
 
-import fastapi
+import uvloop
 import vllm
-from fastapi import APIRouter
-from fastapi.exceptions import RequestValidationError
-from fastapi.middleware.cors import CORSMiddleware
-from fastapi.responses import JSONResponse, Response, StreamingResponse
-from prometheus_client import make_asgi_app
-from starlette.routing import Mount
-from uvicorn import Config as UvicornConfig
-from uvicorn import Server as UvicornServer
-from vllm import envs
-from vllm.engine.arg_utils import AsyncEngineArgs
-from vllm.engine.async_llm_engine import AsyncLLMEngine
-from vllm.entrypoints.logger import RequestLogger
+from vllm.entrypoints.openai.api_server import (
+    build_async_engine_client,
+)
 from vllm.entrypoints.openai.cli_args import make_arg_parser
-from vllm.entrypoints.openai.protocol import (  # noqa: TCH002 # pydantic needs to access these annotations
-    ChatCompletionRequest,
-    ChatCompletionResponse,
-    CompletionRequest,
-    DetokenizeRequest,
-    DetokenizeResponse,
-    EmbeddingRequest,
-    ErrorResponse,
-    TokenizeRequest,
-    TokenizeResponse,
-)
-from vllm.entrypoints.openai.serving_chat import OpenAIServingChat
-from vllm.entrypoints.openai.serving_completion import OpenAIServingCompletion
-from vllm.entrypoints.openai.serving_embedding import OpenAIServingEmbedding
-from vllm.entrypoints.openai.serving_tokenization import (
-    OpenAIServingTokenization,
-)
-from vllm.usage.usage_lib import UsageContext
 from vllm.utils import FlexibleArgumentParser
 
 from .grpc import run_grpc_server
+from .http import run_http_server
 from .logging import init_logger
 from .tgis_utils.args import EnvVarArgumentParser, add_tgis_args, postprocess_tgis_args
 
 if TYPE_CHECKING:
     import argparse
-    from collections.abc import AsyncGenerator
 
-    from vllm.config import ModelConfig
-
-
-TIMEOUT_KEEP_ALIVE = 5  # seconds
-
-openai_serving_chat: OpenAIServingChat
-openai_serving_completion: OpenAIServingCompletion
-openai_serving_embedding: OpenAIServingEmbedding
-openai_serving_tokenization: OpenAIServingTokenization
-
-logger = init_logger(__name__)
-
-_running_tasks: set[asyncio.Task] = set()
+logger = init_logger("vllm-tgis-adapter")
 
 
-router = APIRouter()
+async def start_servers(args: argparse.Namespace) -> None:
+    loop = asyncio.get_running_loop()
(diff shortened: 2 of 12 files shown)


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -199,14 +200,15 @@ def __init__(
...
-        self.config = await self.engine.get_model_config()
+        self.config = await self.client.get_model_config()
 
-        # Swap in the special TGIS stats logger
-        tgis_stats_logger = TGISStatLogger(
-            vllm_stat_logger=self.engine.engine.stat_loggers["prometheus"],
-            max_sequence_len=self.config.max_model_len,
-        )
-        self.engine.engine.stat_loggers["prometheus"] = tgis_stats_logger
+        # FIXME: figure out how to use the TGIS stats logger using the async engine client
```
dtrifiro, line 205: @maxdebayser This is what I think is going to be tricky to get working again

tests/conftest.py
```diff
@@ -1,59 +1,58 @@
...
 
-import asyncio
 import sys
-import threading
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Annotated, TypeVar
 
 import pytest
-import requests
+import pytest_asyncio
```
dtrifiro, line 7: Tests are currently broken. I was experimenting with `pytest_asyncio` to have async fixtures  that can be used to spin up an engine for the whole pytest session, but this is proving to be quite tricky.

src/vllm_tgis_adapter/__main__.py
```diff
@@ -328,41 +61,30 @@ async def run_http_server(
...
+            task.cancel()
 
-    async def run() -> None:
+    async def override_signal_handler() -> None:
         loop = asyncio.get_running_loop()
-
-        http_server_task = loop.create_task(run_http_server(engine, args, model_config))
-        grpc_server_task = loop.create_task(
-            run_grpc_server(
-                engine, args, disable_log_stats=engine_args.disable_log_stats
```
dtrifiro, line 354: Note that I've gotten rid of the `disable_log_stats` argument, as the `_force_log` coroutine is scheduled as part of the `lifespan` attribute of the fastapi server.

pyproject.toml
```diff
@@ -82,7 +83,7 @@ vllm_tgis_adapter = [
 ]
 
 [tool.pytest.ini_options]
-addopts = "-ra -W error::grpc.experimental.ExperimentalApiWarning"
+addopts = "-ra"
```
dtrifiro, line 86: `pytest -W` matches the behaviour of the python cli:

```
The full form of argument is: action:message:category:module:lineno
```

This caused `grpc.experimental.ExperimentalApiWarning` to be raised as an exception when running the healthcheck against the grpc server. We really don't need this.

src/vllm_tgis_adapter/__main__.py
```diff
@@ -328,41 +61,30 @@ async def run_http_server(
...
-        loop.add_signal_handler(signal.SIGTERM, signal_handler)
-
-        await asyncio.gather(grpc_server_task, http_server_task)
-
-    event_loop.run_until_complete(run())
+        while not loop._signal_handlers:  # noqa: SLF001
+            logger.debug("No signal handlers are defined, sleeping")
+            await asyncio.sleep(1)
+
+        logger.debug("Signal handlers have been setup, proceeding with override")
```
dtrifiro, line 78: This trick is required because the signal handlers are set by `vllm.entrypoints.launcher.serve_http`, which are set up at a later stage: we have to wait for them to be set up in order to ovveride them and cleanly shut down the servers.

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -876,25 +878,15 @@ async def ModelInfo(
...
-    # Log memory summary after model is loaded
-    from torch.cuda import memory_summary
-
-    assert isinstance(engine, AsyncLLMEngine)
-    assert isinstance(engine.engine, _AsyncLLMEngine)
-
-    if (device_type := engine.engine.device_config.device.type) == "cuda":
-        logger.info(memory_summary(engine.engine.device_config.device))
-    else:
-        logger.warning("Cannot print device usage for device type: %s", device_type)
```
dtrifiro, line 890: We cannot print the memory summary anymore, as this method is not available from the engine client.

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: We'll fix this in a later PR

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -199,14 +199,16 @@ def __init__(
...
-        )
-        self.engine.engine.stat_loggers["prometheus"] = tgis_stats_logger
+        # FIXME: figure out how to use the TGIS stats logger using the
+        # async engine client
+        # # Swap in the special TGIS stats logger
+        # tgis_stats_logger = TGISStatLogger(
+        #     vllm_stat_logger=self.engine.engine.stat_loggers["prometheus"],
+        #     max_sequence_len=self.config.max_model_len,
+        # )
+        # self.engine.engine.stat_loggers["prometheus"] = tgis_stats_logger
```
njhill, line 211: Can we keep this and wrap it in:
```python
if isinstance(self.client, AsyncLLMEngine):
   ...
else:
    logger.warning("TGIS Metrics currently disabled in decoupled front-end mode, set DISABLE_FRONTEND_MULTIPROCESSING=True to enable")
```

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -950,22 +942,26 @@ async def start_grpc_server(
...
-    disable_log_stats: bool,
-) -> None:
-    assert args is not None
+async def run_grpc_server(args: argparse.Namespace) -> None:
+    from vllm.entrypoints.openai.api_server import AsyncEngineRPCClient
 
-    server = await start_grpc_server(engine, args)
+    rpc_server_port = vllm.envs.VLLM_RPC_PORT
+    logger.info("gRPC server connecting to async engine on port %s", rpc_server_port)
+    async_engine_client = AsyncEngineRPCClient(rpc_server_port)
```
njhill, line 950: I think we need to share the `AsyncEngineClient` from the http server because it might not be an instance of `AsyncEngineRPCClient` .. if running in local mode

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -950,22 +942,26 @@ async def start_grpc_server(
...
-            if not disable_log_stats:
-                await engine.do_log_stats()
     except asyncio.CancelledError:
         print("Gracefully stopping gRPC server")  # noqa: T201
+    except Exception:
+        logger.exception("Engine is dead")
+        # FIXME: handle this case?
+    finally:
         await server.stop(30)  # TODO configurable grace
         await server.wait_for_termination()
```
njhill, line 967: Again not critical but seems like there should be a cleaner way to do this

src/vllm_tgis_adapter/__main__.py
njhill: I'm not sure about the way that async and non-async is mixed here, but it if works that's fine, we can always refactor it a bit later.

njhill (commented):
Thanks @dtrifiro! It's nice that we can remove so much of the duplicated code now

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: The `except Exception` clause was a leftover from an older revision, I'll clean it up in the next push

tests/conftest.py
```diff
@@ -81,13 +76,13 @@ def args(
...
-    monkeysession.setattr(
+    monkeypatch.setattr(
         sys,
         "argv",
         [
             "__main__.py",
-            f"--grpc-port={grpc_server_thread_port}",
-            f"--port={http_server_thread_port}",
+            f"--grpc-port={grpc_server_port}",
+            f"--port={http_server_port}",
```
maxdebayser, line 85: Should we add `"--disable-frontend-multiprocessing"` for now? If there is an exception in the initialization of the server, the tests hang forever even with the timeout.

I noticed this because I sometimes set `NUM_GPUS=0`  and this causes a division by 0 division by 0 error in our tgis adapter code. (But this error is out of scope of this discussion)

maxdebayser: FYI: https://github.com/vllm-project/vllm/issues/7213

src/vllm_tgis_adapter/__main__.py
```diff
@@ -328,41 +61,30 @@ async def run_http_server(
...
-        event_loop = None
-
-    if event_loop is not None and event_loop.is_running():
-        # If the current is instanced by Ray Serve,
-        # there is already a running event loop
-        model_config = event_loop.run_until_complete(engine.get_model_config())
-    else:
-        # When using single vLLM without engine_use_ray
-        model_config = asyncio.run(engine.get_model_config())
+    loop = asyncio.new_event_loop()
```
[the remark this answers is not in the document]
dtrifiro: Done

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: Done

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: Done

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: Cleaned it up

tests/conftest.py
dtrifiro: I cleaned up the tests code a bit, so you might not see that hang anymore.

Regarding `--disable-frontend-multiprocessing`: I parametrized tests so that we test both enabled and disabled.

Tests pass in both cases, although there's an annoying bug at tests cleanup which causes pytest to exit with a non-zero exit code even with all tests passing. I ended up temporarily disabling that test so that we can get this merged asap.

maxdebayser: FYI, https://github.com/vllm-project/vllm/pull/7217 fixes this particular hang.

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: Actually fixed it in this one. See newer versions

dtrifiro:
Parametrization is disabled because of [this](https://github.com/opendatahub-io/vllm-tgis-adapter/pull/84#issuecomment-2273175910) error

PR #84, opened 2024-08-06
tests: enable parametrization

dtrifiro:
- Revert "tests: temporarily disable frontend multiprocessing parametrization"

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -32,7 +32,7 @@ jobs:
         pyv: ["3.11"]
         vllm_version:
           # - "" # skip the pypi version as it will not work on CPU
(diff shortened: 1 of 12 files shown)


dtrifiro:
based on #82

dtrifiro:
Tests pass with parametrization enabled, but `pytest` exists with error:

```
================== 13 passed, 2 skipped, 4 warnings in 25.39s ==================
--- Logging error ---
Fatal Python error: _enter_buffered_busy: could not acquire lock for <_io.BufferedWriter name='<stderr>'> at interpreter shutdown, possibly due to daemon threads
Python runtime state: finalizing (tstate=0x00007f8973376358)

Current thread 0x00007f8973439b80 (most recent call first):
  <no Python frame>
```

This is probably related to the way we spin up vLLM in a loop in a separate thread in the `_servers` fixture.

PR #86, opened 2024-08-07
fix merge_async_iterators usage for vLLM>0.5.4

dtrifiro:
`merge_async_iterators` now has an additional `is_cancelled` kwarg:

https://github.com/vllm-project/vllm/blob/main/vllm/utils.py#L405-L417

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -257,9 +257,18 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
             )
 
         # TODO handle cancellation
-        result_generator: AsyncIterator[tuple[int, RequestOutput]] = (
-            merge_async_iterators(*generators)
-        )
+        result_generator: AsyncIterator[tuple[int, RequestOutput]]
+
+        kwargs = {}
+        if "is_cancelled" in inspect.signature(merge_async_iterators).parameters:
+            # vllm > 0.5.4
+
+            async def is_cancelled() -> bool:
+                return context.cancelled()
+
+            kwargs["is_cancelled"] = is_cancelled
+
+        result_generator = merge_async_iterators(*generators, **kwargs)
 
         resp_options = request.params.response
         responses: list = [None] * request_count


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -257,9 +257,22 @@ async def Generate(
...
+                "'is_cancelled'" not in str(exc)
+            ):
+                raise
+
+            async def is_cancelled() -> bool:
+                return context.cancelled()
+
+            result_generator = merge_async_iterators(
+                *generators, is_cancelled=is_cancelled
+            )
```
dtrifiro, line 275: This does not really do much at this point, since we never call [`context.cancel()`](https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.RpcContext.cancel) , so `context.cancelled()` always return `False` ([docs](https://grpc.github.io/grpc/python/grpc_asyncio.html#grpc.aio.RpcContext.cancelled))

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -257,9 +257,22 @@ async def Generate(
             )
 
         # TODO handle cancellation
-        result_generator: AsyncIterator[tuple[int, RequestOutput]] = (
-            merge_async_iterators(*generators)
-        )
+        result_generator: AsyncIterator[tuple[int, RequestOutput]]
+        try:
+            result_generator = merge_async_iterators(*generators)
+        except TypeError as exc:  # vllm > 0.5.4
```
maxdebayser, line 263: It's a matter of personal preference, but I think using `inspect` would make the code more concise.
```python
if "is_cancelled" in inspect.signature(merge_async_iterators).parameters:
```

maxdebayser (approved):
I've looked a the upstream change that required this change and it makes sense to me.  I've suggested an alternative for the code structure but functionally it does the same thing, so it's fine either way.

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: Agreed, even though this will be removed on the next release, that's cleaner. Done

njhill:
> > I'm not sure that this is correct. It should return False if the client has cancelled/closed their connection.
> 
> I tested this and this is not the case.

I think that's a python grpc bug.

> maybe I'll make a quick upstream PR for that.

Now done and it's merged upstream https://github.com/vllm-project/vllm/pull/7282, so maybe we can revert this change now...

PR #87, opened 2024-08-07
Update comment to be more accurate

njhill:


src/vllm_tgis_adapter/http.py
```diff
@@ -51,7 +51,7 @@ async def run_http_server(
     **uvicorn_kwargs,  # noqa: ANN003
 ) -> None:
     # modified copy of vllm.entrypoints.openai.api_server.run_server that
-    # allows passing of the engine
```
njhill, line 54: The upstream method also allows passing the engine

dtrifiro: That does not seem to be the case: https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/api_server.py#L351-L372

njhill:
OK sorry, I must have been looking at different upstream version at the time I saw this :)

PR #96, opened 2024-08-14
fix: correct parsing of command line bool arg values

tjohnson31415:
## Description

From the [argparse documentation](https://docs.python.org/3/library/argparse.html#argparse-type)
> The [bool()](https://docs.python.org/3/library/functions.html#bool) function is not recommended as a type converter. All it does is convert empty strings to False and non-empty strings to True. This is usually not what is desired.

For example, setting `--disable-prompt-logprobs false` will still parse as `True` for `disable_prompt_logprobs`. Instead, this PR uses a `_bool_from_string` function as the `type` in boolean args to parse a boolean value from the strings contents. The behavior for setting any boolean arg then matches whether it is specified in an env var or as a command-line flag.

## How Has This Been Tested?

I manually experimented with setting the bool flag and inspecting the printed parsed args.

## Merge criteria:

### DIFF
--- a/src/vllm_tgis_adapter/tgis_utils/args.py
+++ b/src/vllm_tgis_adapter/tgis_utils/args.py
@@ -15,6 +15,10 @@ def _to_env_var(arg_name: str) -> str:
     return arg_name.upper().replace("-", "_")
 
 
+def _bool_from_string(val: str) -> bool:
+    return val.lower().strip() == "true" or val == "1"
+
+
 def _switch_action_default(action: argparse.Action) -> None:
     """Switch to using env var fallback for all args."""
     env_val = os.environ.get(_to_env_var(action.dest))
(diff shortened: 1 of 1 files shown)


src/vllm_tgis_adapter/tgis_utils/args.py
```diff
@@ -111,9 +115,11 @@ def add_tgis_args(parser: argparse.ArgumentParser) -> argparse.ArgumentParser:
     # map to tensor_parallel_size
     parser.add_argument("--num-shard", type=int)
     # TODO check boolean behaviour for env vars and defaults
-    parser.add_argument("--output-special-tokens", type=bool, default=False)
     parser.add_argument(
-        "--default-include-stop-seqs", type=bool, default=True
+        "--output-special-tokens", type=_bool_from_string, default=False
```
dtrifiro, line 119: I guess it's a bit odd to use a function as type, but I guess this works.

dtrifiro (approved):

PR #98, opened 2024-08-19
fix: consistent env var parsing for all boolean args in vLLM

tjohnson31415:
Support consistent parsing of boolean arguments defined in vLLM.

## Description

In vLLM, some args use action='store_true':
https://github.com/vllm-project/vllm/blob/df845b2b46c3e30f5bd3e3be286285ed148323fc/vllm/engine/arg_utils.py#L202-L204
and `--enable-chunked-prefill` has a custom arg parsing action called `StoreBoolean`:
https://github.com/vllm-project/vllm/blob/df845b2b46c3e30f5bd3e3be286285ed148323fc/vllm/engine/arg_utils.py#L561-L568

We want `EnvVarArgumentParser` to treat all of these boolean args consistently. The fix in this PR is to inspect the type of the action in addition to `action.type` to detect when we need to parse a boolean from the env var value.

In my testing, I also learned that `argparse` will parse default string values into the correct type for the argument, so there's no need to have special handling for `int` args (which would also imply special handling of `float`, `_bool_from_string`, `nullable_str`, and other typed args).

## How Has This Been Tested?

Tested by hand for the actual vLLM args and also added unit tests of `EnvVarArgumentParser` for the `'store_true'`, `'store_false'`, and `StoreBoolean` action types for boolean args.

## Merge criteria:

### DIFF
--- a/src/vllm_tgis_adapter/tgis_utils/args.py
+++ b/src/vllm_tgis_adapter/tgis_utils/args.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import os
 
+from vllm.engine.arg_utils import StoreBoolean
 from vllm.utils import FlexibleArgumentParser
 
 from vllm_tgis_adapter.grpc.validation import MAX_TOP_N_TOKENS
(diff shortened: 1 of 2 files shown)


njhill (approved):
Thanks @tjohnson31415!

PR #118, opened 2024-09-04
Write to /dev/termination-log on main loop exception

NickLucche:
Addresses https://github.ibm.com/ai-foundation/fmaas-inference-server/issues/722.

## Description

Just like it happens in TGIS https://github.com/IBM/text-generation-inference/blob/9388f02d222c0dab695bea1fb595cacdf08d5467/server/text_generation_server/cli.py#L35, it is useful to have termination logs written to `/dev/termination-log` so that k8s automatically forwards it.

I am not 100% sure this should be brought to vllm-upstream, as I can see them rightfully arguing it should be responsability of the wrapping script, but still I am open to discuss :)

## How Has This Been Tested?
Tested by installing this adapter branch onto a dev pod with vllmv0.5.5 and triggering an injected runtime failure.

Exceptions raising from HTTP server are forwarded just fine to termination logs, but the ones during creation of grpc server currently get overshadowed by a uvicorn one.
Workflow is smt like: grpc server exception awaited-> task cancel sent to all servers -> vllm crashes during exit-coroutine await->exceptions are gathered in `check_for_failed_tasks`->vllm crash only is reported.
Uvicorn stacktrace:

```
INFO 09-04 16:24:09 launcher.py:67] Gracefully stopping http server
INFO:     Shutting down
INFO 09-04 16:24:09 server.py:228] vLLM ZMQ RPC Server was interrupted.
INFO 09-04 16:24:09 multiproc_worker_utils.py:123] Killing local vLLM worker processes
Traceback (most recent call last):
  File "/home/nlucches/work/vllm-tgis-adapter/src/vllm_tgis_adapter/http.py", line 52, in run_http_server
    await shutdown_coro
  File "/home/nlucches/work/vllm-tgis-adapter/.venv/lib/python3.10/site-packages/uvicorn/server.py", line 261, in shutdown
    for server in self.servers:
AttributeError: 'Server' object has no attribute 'servers'. Did you mean: 'serve'?

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/nlucches/.local/share/uv/python/cpython-3.10.14-linux-x86_64-gnu/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/nlucches/.local/share/uv/python/cpython-3.10.14-linux-x86_64-gnu/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/nlucches/work/vllm-tgis-adapter/src/vllm_tgis_adapter/__main__.py", line 98, in <module>
    raise e
  File "/home/nlucches/work/vllm-tgis-adapter/src/vllm_tgis_adapter/__main__.py", line 94, in <module>
    loop.run_until_complete(start_servers(args))
  File "uvloop/loop.pyx", line 1517, in uvloop.loop.Loop.run_until_complete
  File "/home/nlucches/work/vllm-tgis-adapter/src/vllm_tgis_adapter/__main__.py", line 72, in start_servers
    check_for_failed_tasks(tasks)
  File "/home/nlucches/work/vllm-tgis-adapter/src/vllm_tgis_adapter/utils.py", line 27, in check_for_failed_tasks
    raise RuntimeError(f"task={name} ({coro_name}) exception={str(exc)}") from exc
RuntimeError: task=http_server (run_http_server) exception='Server' object has no attribute 'servers'

```
 @joerunde 
## Merge criteria:

### DIFF
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -2,6 +2,8 @@ from __future__ import annotations
 
 import asyncio
 import contextlib
+import os
+import traceback
 from concurrent.futures import FIRST_COMPLETED
 from typing import TYPE_CHECKING
 
@@ -17,7 +19,7 @@ from .grpc import run_grpc_server
 from .http import run_http_server
 from .logging import init_logger
 from .tgis_utils.args import EnvVarArgumentParser, add_tgis_args, postprocess_tgis_args
-from .utils import check_for_failed_tasks
+from .utils import check_for_failed_tasks, write_termination_log
 
 if TYPE_CHECKING:
     import argparse
@@ -43,10 +45,21 @@ async def start_servers(args: argparse.Namespace) -> None:
         )
         tasks.append(grpc_server_task)
 
+        runtime_error = None
         with contextlib.suppress(asyncio.CancelledError):
             # Both server tasks will exit normally on shutdown, so we await
             # FIRST_COMPLETED to catch either one shutting down.
             await asyncio.wait(tasks, return_when=FIRST_COMPLETED)
+            if engine and engine.errored and not engine.is_running:
+                # both servers shut down when an engine error
+                # is detected, with task done and exception handled
+                # here we just notify of that error and let servers be
+                runtime_error = RuntimeError(
+                    "AsyncEngineClient error detected,this may be caused by an \
+                        unexpected error in serving a request. \
+                        Please check the logs for more details."
+                )
+
         # Once either server shuts down, cancel the other
         for task in tasks:
             task.cancel()
@@ -55,6 +68,22 @@ async def start_servers(args: argparse.Namespace) -> None:
         await asyncio.wait(tasks)
 
         check_for_failed_tasks(tasks)
+        if runtime_error:
+            raise runtime_error
+
+
+def run_and_catch_termination_cause(
+    loop: asyncio.AbstractEventLoop, task: asyncio.Task
+) -> None:
+    try:
+        loop.run_until_complete(task)
+    except Exception:
+        # Report the first exception as cause of termination
+        msg = traceback.format_exc()
+        write_termination_log(
+            msg, os.getenv("TERMINATION_LOG_DIR", "/dev/termination-log")
+        )
+        raise
 
 
 if __name__ == "__main__":
@@ -75,4 +104,5 @@ if __name__ == "__main__":
 
     asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
     loop = asyncio.new_event_loop()
-    loop.run_until_complete(start_servers(args))
+    task = loop.create_task(start_servers(args))
+    run_and_catch_termination_cause(loop, task)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -11,7 +11,7 @@ import vllm
 from vllm.entrypoints.openai.cli_args import make_arg_parser
 from vllm.utils import FlexibleArgumentParser
 
-from vllm_tgis_adapter.__main__ import start_servers
+from vllm_tgis_adapter.__main__ import run_and_catch_termination_cause, start_servers
 from vllm_tgis_adapter.grpc.grpc_server import TextGenerationService
 from vllm_tgis_adapter.healthcheck import health_check
 from vllm_tgis_adapter.tgis_utils.args import (
@@ -69,6 +69,11 @@ def disable_frontend_multiprocessing(request):
 
 
 @pytest.fixture
+def server_args(request: pytest.FixtureRequest):
+    return request.param if hasattr(request, "param") else []
+
+
+@pytest.fixture
 def args(  # noqa: PLR0913
     request: pytest.FixtureRequest,
     monkeypatch,
@@ -76,11 +81,13 @@ def args(  # noqa: PLR0913
     http_server_port: ArgFixture[int],
     lora_available: ArgFixture[bool],
     disable_frontend_multiprocessing,
+    server_args: ArgFixture[list[str]],
 ) -> argparse.Namespace:
     """Return parsed CLI arguments for the adapter/vLLM."""
     # avoid parsing pytest arguments as vllm/vllm_tgis_adapter arguments
 
-    extra_args: list[str] = []
+    # Extra server init flags
+    extra_args: list[str] = [*server_args]
     if lora_available:
         name = request.getfixturevalue("lora_adapter_name")
         path = request.getfixturevalue("lora_adapter_path")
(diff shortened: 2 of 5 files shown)


joerunde:
Thanks @NickLucche!

> I am not 100% sure this should be brought to vllm-upstream, as I can see them rightfully arguing it should be responsability of the wrapping script, but still I am open to discuss :)

I think they should be open to it, since vLLM provides an official image that's designed to be ready to deploy. I think it's worth proposing, at least. But from glancing at the diff here it does look like we'd probably still need to implement this functionality separately here since we'd want to catch errors from the grpc server as well.

Regarding testing these changes, it'd be nice to see some examples from misconfiguring the server. I think that's the primary use case here, helping people debug what mistake they made with their deployment to get it up and running. Some common things you can do are:
- Enable lora on a model format that doesn't support it (the default facebook/opt-125m will fail)
- Supply lora modules but don't enable lora
- Set a context length that's longer than what the model supports
- Deploy an unsupported model architecture (all encoder-decoder models should fail)
- Deploy a model that's too large to fit in memory

As is, I think most of those should fail when creating the engine, so the exception will be raised before the `check_for_failed_tasks` check and I think they'd format nicely. 

For other errors that crash the server at runtime, I think we'd want to try to output the root cause and not the wrapping like `RuntimeError: task=http_server (run_http_server) exception='{blah}'`. There are some edge cases that you can use to crash the server, but the easiest way is probably to go and edit the `LLMEngine` code locally to raise an error when it goes to run the model.

Also, it looks like you're running into issues with the formatter, so you may want to try to installing and enabling pre-commit to ensure everything is formatted correctly

src/vllm_tgis_adapter/__main__.py
```diff
@@ -75,4 +75,9 @@ async def start_servers(args: argparse.Namespace) -> None:
 
     asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
     loop = asyncio.new_event_loop()
-    loop.run_until_complete(start_servers(args))
+    try:
+        loop.run_until_complete(start_servers(args))
+    except RuntimeError as e:
+        # Report the first exception as cause of termination
+        write_termination_log(str(e), "/dev/termination-log")
+        raise e
```
dtrifiro, line 83:
```suggestion
        raise
```

NickLucche:
I added a way to test for some of those cases you mentioned; getting all cli errors is a bit tricky though as they may happen when the engine is created (in vllm space) but still in a separate serrver process.

tests/conftest.py
```diff
@@ -58,13 +58,14 @@ def lora_adapter_path(request: pytest.FixtureRequest) -> str:
...
+    # params=[
         # pytest.param(True, id="disable-frontend-multiprocessing=True"),
-        pytest.param(False, id="disable-frontend-multiprocessing=False"),
-    ]
+        # pytest.param(False, id="disable-frontend-multiprocessing=False"),
+    # ]
 )
 def disable_frontend_multiprocessing(request):
     """Enable or disable the frontend-multiprocessing feature."""
+    return False
```
dtrifiro: Why were the parameters commented and the fixture modified to return an hardcoded `False`?

tests/conftest.py
```diff
@@ -80,7 +81,8 @@ def args(  # noqa: PLR0913
     """Return parsed CLI arguments for the adapter/vLLM."""
     # avoid parsing pytest arguments as vllm/vllm_tgis_adapter arguments
 
-    extra_args: list[str] = []
```
dtrifiro, line 83: What's this for? Unless I'm not seeing something, I don't think this should be changed

tests/conftest.py
NickLucche: my bad fixed in new commit

tests/conftest.py
```diff
@@ -195,7 +201,7 @@ def target():
 
         t.join()
 
-    # rorkaround: Instantiating the TGISStatLogger multiple times creates
+    # workaround: Instantiating the TGISStatLogger multiple times creates
```
dtrifiro, line 204: :100:

tests/test_termination_log.py
```diff
@@ -0,0 +1,44 @@
...
+
+    These errors should be properly reported in the termination log.
+
+    """
+    if lora_available and args.enable_lora:
+        pytest.skip("This test requires a non-lora supported device to run")
+
+    # execute failing server fixture
+    with pytest.raises(TaskFailedError):
+        _ = request.getfixturevalue("_create_server")
```
dtrifiro, line 41: Can't we just use `_servers` here instead of introducing `_create_server` ?

dtrifiro (approved):

tests/test_termination_log.py
NickLucche: absolutely, my bad initial draft was different

src/vllm_tgis_adapter/__main__.py
```diff
@@ -55,6 +68,22 @@ async def start_servers(args: argparse.Namespace) -> None:
...
+    loop: asyncio.AbstractEventLoop, task: asyncio.Task
+) -> None:
+    try:
+        loop.run_until_complete(task)
+    except Exception:
+        # Report the first exception as cause of termination
+        msg = traceback.format_exc()
+        write_termination_log(
+            msg, os.getenv("TERMINATION_LOG_DIR", "/dev/termination-log")
+        )
```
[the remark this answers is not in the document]
NickLucche: I can revert that to str, but most server start exceptions will make even less sense

NickLucche:
@dtrifiro it seems unrelated to this PR but my tests are sometimes hanging on local when checking out master `7d94eae` here (`[build(deps): bump mypy from 1.11.1 to 1.11.2])` with:

```bash
INFO 09-13 16:45:09 logs.py:155] generate{input=[b'The answer to life the universe ...', b'Medicinal herbs '] prefix_id= correlation_id=None adapter_id= input_chars=[66] params=stopping { max_new_tokens: 10 } tokenization_time=2.47ms queue_time=0.97ms inference_time=2465.34ms time_per_token=246.53ms total_time=2468.77ms input_toks=5}: Sub-request 1 from batch of 2 generated 10 tokens before MAX_TOKENS, output 10 chars: b'          '
.Gracefully stopping gRPC server
INFO 09-13 16:45:09 launcher.py:67] Gracefully stopping http server
INFO:     Shutting down
INFO:     Waiting for application shutdown.
INFO:     Application shutdown complete.
INFO 09-13 16:45:09 server.py:228] vLLM ZMQ RPC Server was interrupted.
INFO 09-13 16:45:09 async_llm_engine.py:60] Engine is gracefully shutting down.
INFO 09-13 16:45:09 multiproc_worker_utils.py:136] Terminating local vLLM worker processes

```

Ran them with `python -m pytest -s tests` 🤔

src/vllm_tgis_adapter/__main__.py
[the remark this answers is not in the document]
dtrifiro: Well, I guess it will be truncated, let's get this merged.

PR #128, opened 2024-09-16
grpc_server: use x-correlation-id as request-id when possible

dtrifiro:
> Currently, the vLLM OpenTelemetry span exports the request_id, but inspecting the WatsonX trace in Instana can be confusing due to the presence of multiple IDs.
Using the x-correlation-id as the request_id would help make the trace more coherent.

https://issues.redhat.com/browse/RHOAIENG-12927

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -518,8 +518,17 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
         return response
 
     @staticmethod
-    def request_id(context: ServicerContext) -> str:  # noqa:  ARG004
-        return uuid.uuid4().hex
+    def request_id(context: ServicerContext) -> str:
+        metadata = context.invocation_metadata()
+        if not metadata:
+            return uuid.uuid4().hex
+
+        correlation_id = dict(metadata).get("x-correlation-id")
+
+        if not correlation_id:
+            return uuid.uuid4().hex
+
+        return correlation_id
 
     async def _validate_and_convert_params(
         self,
--- a/tests/test_grpc_server.py
+++ b/tests/test_grpc_server.py
@@ -55,3 +55,30 @@ def test_lora_request(grpc_client, lora_adapter_name):
     response = grpc_client.make_request("hello", adapter_id=lora_adapter_name)
 
     assert response.text
+
+
+def test_request_id(grpc_client, mocker):
+    from vllm_tgis_adapter.grpc.grpc_server import TextGenerationService, uuid
+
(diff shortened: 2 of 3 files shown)


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -512,8 +512,17 @@ def _convert_output(  # noqa: PLR0913
...
 
     @staticmethod
-    def request_id(context: ServicerContext) -> str:  # noqa:  ARG004
-        return uuid.uuid4().hex
+    def request_id(context: ServicerContext) -> str:
+        metadata = context.invocation_metadata()
+        if not metadata:
+            return uuid.uuid4().hex
+
+        correlation_id = dict(metadata).get("x-correlation-id")
```
ronensc, line 520: We have the same "magic string" in `log.py`:
https://github.com/opendatahub-io/vllm-tgis-adapter/blob/20db20b3ba4b184fde441b6ea5a9df6ee6894850/src/vllm_tgis_adapter/tgis_utils/logs.py#L137

Would it be worthwhile to extract it into a constant?

ronensc (approved):

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -512,8 +512,17 @@ def _convert_output(  # noqa: PLR0913
...
+
+        if not correlation_id:
+            return uuid.uuid4().hex
+
+        return correlation_id
```
ronensc, line 525: Is the compact version clearer to read?
```suggestion
         metadata = dict(context.invocation_metadata() or {})
         return metadata.get("x-correlation-id", uuid.uuid4().hex)
```

dtrifiro: Compared to the compact version, we avoid evaluating `uuid.uuid4()` when we have the correlation id with the current code, I'm fine either way tbh

src/vllm_tgis_adapter/grpc/grpc_server.py
dtrifiro: Probably not

src/vllm_tgis_adapter/grpc/grpc_server.py
ronensc: You're right, this can be addressed by replacing the return statement with:
```python
return metadata.get("x-correlation-id") or uuid.uuid4().hex
```
I was initially triggered by the repeated use of `uuid4()`, aiming to follow the DRY principle, but I'm okay with whichever approach you prefer.

dtrifiro:
Waiting on #138 before merging

NickLucche (approved):

PR #129, opened 2024-09-16
do not run codecov upload step on dependabot PR

NickLucche:
Addressing https://issues.redhat.com/browse/RHOAIENG-12847. Dependabot PRs can't auto-merge because codecov update will fail, as PRs from forks are not allowed to access secrets (to avoid exposure).

## Description

Have a separate PR workflow that checks whether it's coming from dependabot. Should be safe from secrete leaking as other steps (the ones shared with all contributors) do NOT have the codecov secret injected.


dtrifiro:
Thinking about it, it probably makes more sense to just skip the code coverage upload  step for dependabot PRs, since they won't involve code changes.

NickLucche:
In light of https://docs.codecov.com/docs/codecov-tokens#uploading-without-a-token, I think we can keep it simple as you suggested

.github/workflows/tests.yaml
```diff
@@ -75,6 +75,7 @@ jobs:
 
       - name: Upload coverage report
         uses: codecov/codecov-action@v4
+        if: github.actor != 'dependabot[bot]'
```
dtrifiro, line 78: :100:

dtrifiro:
Looks great, can you update the PR title aswell? Thanks!

dtrifiro (approved):

PR #130, opened 2024-09-16
Propagate cancellation of gRPC requests

njhill:
@dtrifiro added a partial change for this while ago which was reverted to maintain compatibility with older vLLM version.

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -233,8 +233,8 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
         start_time = time.time()
         service_metrics.count_generate_request(len(request.requests))
         request_id = self.request_id(context)
-        adapter_kwargs = await self._validate_adapters(request, context)
-        tokenizer = await self._get_tokenizer(adapter_kwargs)
+        kwargs = await self._validate_adapters(request, context)
+        tokenizer = await self._get_tokenizer(kwargs)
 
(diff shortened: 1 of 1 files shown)


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -253,7 +254,7 @@ async def Generate(
                 prompt=req.text,
                 prompt_token_ids=input_ids,
             )
-            kwargs = {}
+            kwargs = adapter_kwargs
```
dtrifiro, line 257: I guess we can just get rid of this>

dtrifiro (approved):

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -26,6 +26,7 @@
     extract_trace_headers,
     log_tracing_disabled_warning,
 )
+from vllm.utils import iterate_with_cancellation
```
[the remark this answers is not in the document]
njhill: Missing from where? it's in v0.5.5...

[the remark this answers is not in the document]
dtrifiro: You're right, I'm sorry, I thought it was in the adapter utils.

src/vllm_tgis_adapter/grpc/grpc_server.py
njhill: You mean get rid of separate `adapter_kwargs` var?

dtrifiro: yeah

PR #137, opened 2024-09-17
Exploit vLLM options to return deltas/final-output only

njhill:
Nontrivial performance benefit, particularly when running with decoupled front-end process.

These changes require vLLM >= `0.6.1.post2`

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -17,10 +17,10 @@ from grpc import StatusCode, aio
 from grpc._cython.cygrpc import AbortError
 from grpc_health.v1 import health, health_pb2, health_pb2_grpc
 from grpc_reflection.v1alpha import reflection
-from vllm import SamplingParams
 from vllm.engine.async_llm_engine import AsyncLLMEngine
 from vllm.entrypoints.openai.serving_completion import merge_async_iterators
 from vllm.inputs import LLMInputs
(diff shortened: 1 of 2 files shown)


njhill:
@dtrifiro if necessary I could update these changes to work both before and after v0.6.2.post2 (like you've done for other things)

njhill:
Now that vllm 0.6.2 is released, @dtrifiro agreed that there's no need for the aforementioned backwards compatibility changes, so this PR should now be ready to merge.

dtrifiro:
@njhill it seems this broke streaming generation

``` E                   grpc._channel._MultiThreadedRendezvous: <_MultiThreadedRendezvous of RPC that terminated with:
E                   	status = StatusCode.UNKNOWN
E                   	details = "Unexpected <class 'TypeError'>: object of type 'NoneType' has no len()"
E                   	debug_error_string = "UNKNOWN:Error received from peer ipv6:%5B::1%5D:45943 {created_time:"2024-09-27T00:19:45.777203684+00:00", grpc_status:2, grpc_message:"Unexpected <class \'TypeError\'>: object of type \'NoneType\' has no len()"}"
E                   >
```

njhill:
@dtrifiro this one is now fixed and should be working

njhill:
Remaining test failures appeared to be unrelated to this PR.

Probably due to other upstream changes, and we still need to investigate/address of course.

njhill:
Yeah looks like those are due to an upstream bug with CPU backend https://github.com/vllm-project/vllm/issues/9024.

dtrifiro (approved):

PR #136, opened 2024-09-17
http_server: compatibility fixes for vllm>0.6.1.post1

dtrifiro:
- https://github.com/vllm-project/vllm/pull/8492 renamed `init_app` to `init_app_state`
- gha: bump vllm tag to `v0.6.1.post2`

https://issues.redhat.com/browse/RHOAIENG-12997

PR #138, opened 2024-09-23
Fix calls to defunct AsyncEngineClient

NickLucche:
Address breaking changes in upstream main (v0.6.2?) https://github.com/vllm-project/vllm/pull/8673 and https://github.com/vllm-project/vllm/pull/8157. The latter is yet another bulky change to the rpc client-server interaction: I couldn't spot differences in functionality and tests are fine on our side, but I could really use some help to double check that what's been implemented upstream is still fully compatible with the use being made of here.

I branched off https://github.com/opendatahub-io/vllm-tgis-adapter/pull/136, so we should be merging that PR first.

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -262,7 +262,7 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
                 log_tracing_disabled_warning()
             generators.append(
                 self.engine.generate(
-                    inputs,
+                    inputs=inputs,
                     sampling_params=sampling_params,
                     request_id=f"{request_id}-{i}",
                     **adapter_kwargs,
@@ -363,7 +363,7 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
         result_generator = self.engine.generate(
             # prompt is supplied for observability, the text is not
             # re-tokenized when `prompt_token_ids` is supplied
-            inputs,
+            inputs=inputs,
             sampling_params=sampling_params,
             request_id=request_id,
             **adapter_kwargs,
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -65,7 +65,12 @@ if TYPE_CHECKING:
     from transformers import PreTrainedTokenizer
     from vllm import CompletionOutput, RequestOutput
     from vllm.config import ModelConfig
-    from vllm.engine.protocol import EngineClient
+
+    try:
+        from vllm.engine.protocol import EngineClient
+    except ImportError:
+        # fallback for versions <=v0.6.1.post2
+        from vllm.engine.protocol import AsyncEngineClient as EngineClient
     from vllm.lora.request import LoRARequest
     from vllm.sequence import Logprob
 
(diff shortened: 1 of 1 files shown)


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -262,7 +262,7 @@ async def Generate(
                 log_tracing_disabled_warning()
             generators.append(
                 self.engine.generate(
-                    inputs=inputs,
+                    inputs,
```
dtrifiro, line 265: I'd rather not remove this, I prefer explicitly specifying the names of the input, as this has caused issues in the past

dtrifiro (approved):
Looks good, one minor nit

src/vllm_tgis_adapter/grpc/grpc_server.py
NickLucche: can I move `inputs` to `prompt` and assume `>v0.6.1.post2` or do I have to be backward compatible?

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -177,12 +177,12 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
 
     def __init__(
         self,
-        engine: AsyncEngineClient | AsyncLLMEngine,
+        engine: EngineClient | AsyncLLMEngine,
```
njhill, line 180: I'm not sure that the union should be needed here.. `EngineClient` is a `Protocol` and `AsyncLLMEngine` implements that protocol.. so only the former should be needed I think?

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -65,7 +65,7 @@
     from transformers import PreTrainedTokenizer
     from vllm import CompletionOutput, RequestOutput
     from vllm.config import ModelConfig
-    from vllm.engine.protocol import AsyncEngineClient
+    from vllm.engine.protocol import EngineClient
```
njhill, line 68: Since this was renamed in the latest changes, won't this break usage of the adapter with existing releases (which I thought @dtrifiro tries to maintain)?

joerunde:
```suggestion
    try:
        from vllm.engine.protocol import EngineClient
    except ImportError:
        from vllm.engine.protocol import AsyncEngineClient as EngineClient
```
Maybe this is simple to alias like so?

src/vllm_tgis_adapter/grpc/grpc_server.py
joerunde: Ah, sad this wasn't done in a non-breaking way in vllm: https://github.com/vllm-project/vllm/pull/8673/files#diff-e8f334a55697b72398acbbed4d280bcb3e5eb6c2e66fc81ccde477818b4b0351R38. Maybe we can simply comment here that the kwarg name is removed due to this change, and then in later releases add in `prompt=`?

Might also be worth changing the name of the `inputs` variable here to `prompt`

njhill: We are reverting upstream for now https://github.com/vllm-project/vllm/pull/8750.

NickLucche: I guess it could've been done in a less abrupt way upstream by warning about deprecation of `inputs` while adding both `inputs` and `prompts` for some time..

src/vllm_tgis_adapter/grpc/grpc_server.py
NickLucche: yep my bad, good one!

joerunde:
Thanks @NickLucche!

PR #147, opened 2024-09-30
gha: add ccache

dtrifiro:
should speed up CI a tiny bit

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -53,6 +53,7 @@ jobs:
         run: |
           sudo apt update
           sudo apt install --no-install-recommends -y \
+            ccache \
             libnuma-dev libdnnl-dev opencl-dev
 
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -79,6 +80,13 @@ jobs:
       - name: Lint code and check dependencies
         run: nox --envdir ~/.nox --reuse-venv=yes -v -s lint-${{ matrix.pyv }}
 
+      - name: Cache ccache cache directory
+        id: cache-ccache
+        uses: actions/cache@v4
+        with:
+          path: /home/runner/.cache/ccache
+          key: ${{ runner.os }}
```
dtrifiro, line 88: This key is fairly generic, we might want to tweak it.

Another issue that came up with this PR is that there's contention for the cache because all three jobs use the same key, so we might want to tweak that aswell.

PR #156, opened 2024-10-05
Fail startup with root-cause exception

njhill:
If the gRPC server startup fails, the http server task can also fail for some other reason when cancelled. The current logic looks for an arbitrary failed task after this and raises an exception based on that.

We want to do this based on the root cause exception not the secondary one from the other task's cancellation. So that the root cause is not lost.


dtrifiro:
Can you solve the merge conflicts?

njhill:
@dtrifiro now resolved!

dtrifiro (approved):

PR #160, opened 2024-10-09
fix linter complaints, cleanup gha caching

dtrifiro:


.github/workflows/tests.yaml
```diff
@@ -73,7 +74,7 @@ jobs:
         uses: actions/cache@v4
         with:
           path: /home/runner/.nox
-          key: ${{ runner.os }}-${{ matrix.pyv }}-${{ hashFiles('noxfile.py') }}
+          key: ${{ runner.os }}-python-${{steps.setup-python.outputs.python-version }}-${{ hashFiles('noxfile.py') }}
```
dtrifiro, line 77: Some CI failures seemed to be caused by the fact that the cache was create for python `3.12.6` but actions were setting up the newly released `3.12.7`. 

Let's use the correct key by using the setup-python action's [outputs](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#outputs)

PR #161, opened 2024-10-10
:sparkles: invoke caikit -> peft conversion at load time

joerunde:
## Description
This creates new tempdirs and invokes the caikit to peft conversion utility at prompt adapter load time when a `decoder.pt` format model is detected.

## How Has This Been Tested?
Loaded up onto a dev pod and tested with the `bloom_sentiment_1` fixture that was added with the conversion utility. Seems to work fine!

Need to add a unit test for this specifically, still have to get nox up and running locally. Opening this PR now for feedback.

## Merge criteria:


src/vllm_tgis_adapter/grpc/adapters.py
```diff
@@ -134,17 +139,35 @@ def _load_adapter_config_from_file(adapter_id: str, adapter_path: str) -> dict:
             adapter_id, "directory does not exist"
         )
 
+    # 🌶️🌶️🌶️ Check for caikit-style adapters first
+    if Path(adapter_path).exists() and (Path(adapter_path) / "decoder.pt").exists():
```
dtrifiro, line 143:
```suggestion
    if (Path(adapter_path) / "decoder.pt").exists():
```

Wouldn't this be enough?

dtrifiro (approved):

src/vllm_tgis_adapter/grpc/adapters.py
joerunde: you right ❤️

PR #162, opened 2024-10-11
🐛 handle MistralTokenizer special case

prashantgupta24:
## Description
Right now, [MistralTokenizer](https://github.com/vllm-project/vllm/blob/v0.6.2/vllm/transformers_utils/tokenizers/mistral.py#L43) doesn't support the `encode_plus` option. We use the `encode` option for now, and throw an error to the user if `return_offset` is requested.

Also, `mistral` models do NOT select `mistral` as the tokenizer by default. `TOKENIZER_MODE=mistral` has to be set in order to do that.

## Merge criteria:

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -23,6 +23,7 @@ from vllm.tracing import (
     log_tracing_disabled_warning,
 )
 from vllm.transformers_utils.tokenizer import AnyTokenizer  # noqa: TCH002
+from vllm.transformers_utils.tokenizers.mistral import MistralTokenizer
 from vllm.utils import iterate_with_cancellation
 
 from vllm_tgis_adapter.logging import init_logger
@@ -855,18 +856,25 @@ class TextGenerationService(generation_pb2_grpc.GenerationServiceServicer):
         tokenizer = await self._get_tokenizer(adapter_kwargs)
 
         responses: list[TokenizeResponse] = []
+        is_mistral_tokenizer = isinstance(tokenizer, MistralTokenizer)
 
         # TODO: maybe parallelize, also move convert_ids_to_tokens into the
         # other threads
         for req in request.requests:
-            batch_encoding = tokenizer.encode_plus(
-                text=req.text,
-                return_offsets_mapping=request.return_offsets,
-                add_special_tokens=ADD_SPECIAL_TOKENS,
-            )
+            if is_mistral_tokenizer:
+                token_ids = tokenizer.encode(
+                    prompt=req.text,
+                )
+            else:
+                batch_encoding = tokenizer.encode_plus(
+                    text=req.text,
+                    return_offsets_mapping=request.return_offsets,
+                    add_special_tokens=ADD_SPECIAL_TOKENS,
+                )
 
-            # Tokenize the input text
-            token_ids = batch_encoding.input_ids
+                # Tokenize the input text
+                token_ids = batch_encoding.input_ids
+
             token_count = len(token_ids)
 
             if 0 < request.truncate_input_tokens < token_count:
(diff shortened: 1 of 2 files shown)


src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -861,18 +856,25 @@ async def Tokenize(
...
 
         # TODO: maybe parallelize, also move convert_ids_to_tokens into the
         # other threads
         for req in request.requests:
-            batch_encoding = tokenizer.encode_plus(
-                text=req.text,
-                return_offsets_mapping=request.return_offsets,
-                add_special_tokens=ADD_SPECIAL_TOKENS,
-            )
+            if is_mistral_tokenizer:
```
prashantgupta24, line 864: I guess I could also check if the tokenizer has the `encode_plus` attribute to keep things generic

njhill: @prashantgupta24 yes this would probably be better, but I think this is ok too.

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -883,13 +885,19 @@ async def Tokenize(
...
-                    for start, end in batch_encoding.offset_mapping
-                    if start is not None and end is not None
-                ]
-                # Truncate offset list if request.truncate_input_tokens
-                offsets = offsets[-token_count:]
+                if is_mistral_tokenizer:
+                    logger.warning(
+                        "Mistral tokenizer doesn't support "
+                        "return_offsets at the moment. "
+                    )
```
njhill, line 892: I actually think it would be best to raise a `ValueError` here with a message like "this model's tokenizer does not the return_offsets option".

njhill (commented):
Thanks @prashantgupta24, looks great just a couple of comments.

src/vllm_tgis_adapter/grpc/grpc_server.py
prashantgupta24: done

src/vllm_tgis_adapter/grpc/grpc_server.py
prashantgupta24: Doing this now

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -883,6 +885,11 @@ async def Tokenize(
             offsets = None
 
             if request.return_offsets:
+                if is_mistral_tokenizer:
+                    raise ValueError(
+                        "Mistral tokenizer doesn't support "
+                        "return_offsets at the moment. "
+                    )
```
njhill, line 892: Another small thought - probably better to move this above the tokenizer.encode above .. i.e. have an `if request.return_offsets: raise ValueError...` in that same `if mistral_tokenizer` block (so that it doesn't unnecessarily perform the tokenization in this case)

prashantgupta24: good idea

src/vllm_tgis_adapter/grpc/grpc_server.py
prashantgupta24: Done

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -865,14 +859,25 @@ async def Tokenize(
...
-            batch_encoding = tokenizer.encode_plus(
-                text=req.text,
-                return_offsets_mapping=request.return_offsets,
-                add_special_tokens=ADD_SPECIAL_TOKENS,
-            )
+            if not hasattr(tokenizer, "encode_plus"):
+                if request.return_offsets:
+                    raise ValueError(
+                        f"{type(tokenizer)} doesn't support "
+                        "return_offsets at the moment. "
```
njhill, line 866: suggested wording adjustment
```suggestion
                        "the return_offsets option"
```

njhill (commented):
One last nit

njhill (approved):
Thanks @prashantgupta24!

PR #191, opened 2024-12-06
:loud_sound: overhaul logging on engine

joerunde:
## Description
This PR addresses the problem that logging is inconsistent between the http server and grpc server. When users make /completions/chat calls, they don't get any of the formatted log messages that we've wrapped the grpc invocations with.
Additionally, our existing logs for the grpc server include parts of the request and response, which is often inappropriate for logging in production settings governed by privacy laws and organization policy. This PR also strips that data out of the logging.

To do this, the `EngineClient` that is created at startup has its `generate` method wrapped via `functools.wrap`, and the wrapper logs both request and response information. This ensures that the http and grpc servers log everything the same way, and that we get both request and response logging where we only had response logging before. 

A small challenge here is that we support adding a "correlation id" to logs that should be populated from the `x-correlation-id` header, and at the engine level we have no context about either the http or grpc request to retrieve the header. To get around this, we add a little blackboard cache where we can map request ids to correlation ids. Then in the grpc handlers and in an http middleware method, we fetch the `x-correlation-id` header and cache it with the request id.

An example of the old logging:
```
NFO 12-02 05:36:30 logs.py:155] generate{input=[b'what is the distance between par...'] prefix_id= correlation_id=text-infernece-test-call-31307 adapter_id= input_chars=[53] params=stopping { max_new_tokens: 20 min_new_tokens: 20 } decoding { repetition_penalty: 2.0 } tokenization_time=1.12ms queue_time=0.41ms inference_time=381.83ms time_per_token=19.09ms total_time=383.36ms input_toks=11}: Request generated 20 tokens before MAX_TOKENS, output 101 chars: b' The average flight time from Pa...'
```

And an example of the new logging:
```
INFO 12-06 21:53:18 logs.py:135] Processing request: {request_id=cmpl-foo-0, correlation_id=foo, adapter_id=None,  input_tokens=2,params=SamplingParams(n=1, presence_penalty=0.0, frequency_penalty=0.0, repetition_penalty=1.0, temperature=1.0, top_p=1.0, top_k=-1, min_p=0.0, seed=None, stop=[], stop_token_ids=[], bad_words=[], include_stop_str_in_output=False, ignore_eos=False, max_tokens=16, min_tokens=0, logprobs=None, prompt_logprobs=None, skip_special_tokens=True, spaces_between_special_tokens=True, truncate_prompt_tokens=None, guided_decoding=None)}
INFO 12-06 21:53:18 logs.py:182] Finished processing request: {request_id=cmpl-foo-0, correlation_id=foo}. Timing info: {queue_time=0.65ms, inference_time=23.25ms, time_per_token=2.33ms, total_time=24.51ms}. Generated 10 tokens before finish reason: stop, output 28 chars
```

A drawback is that this changes the format of the logs from the old TGIS format, but... 🤷 

Will require https://github.com/vllm-project/vllm/pull/10968 for the correlation id to be propagated from the http server

## How Has This Been Tested?

Just installing the changes and running a server, and sending both grpc and http requests and looking at the logs.

## Merge criteria:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -33,6 +33,7 @@ dependencies = [
   "grpcio-reflection==1.62.2",
   "accelerate==0.34.2",
   "hf-transfer==0.1.8",
+  "cachetools~=5.5",
   # additional dependencies for OpenTelemetry tracing
   "opentelemetry-sdk>=1.26.0,<1.28.0",
   "opentelemetry-api>=1.26.0,<1.28.0",
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -15,6 +15,8 @@ from vllm.entrypoints.openai.api_server import (
 from vllm.entrypoints.openai.cli_args import make_arg_parser
 from vllm.utils import FlexibleArgumentParser
 
+from vllm_tgis_adapter.tgis_utils.logs import add_logging_wrappers
+
 from .grpc import run_grpc_server
 from .http import run_http_server
 from .logging import init_logger
@@ -32,6 +34,8 @@ async def start_servers(args: argparse.Namespace) -> None:
 
     tasks: list[asyncio.Task] = []
     async with build_async_engine_client(args) as engine:
+        add_logging_wrappers(engine)
+
         http_server_task = loop.create_task(
             run_http_server(args, engine),
             name="http_server",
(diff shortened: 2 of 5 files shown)


dtrifiro:
This looks good. I just have a minor concern regarding the introduction of `cachetools`, how about using `functools.lru_cache` without introduction a new dependency?

joerunde:
@dtrifiro I'm not sure that `functools.lru_cache` does what we need though since it memoizes a function, and what we need here is a bounded cache that we can explicitly add items to. 

A different approach here to avoid dealing with a correlation id cache would be to keep the logging at the service level- we could inject middleware onto the http app that logs the request and response. The issue there is that then we have no access to the engine-level metrics to fetch the queue time, inference time, and time per token which have been pretty useful for debugging issues in prod

dtrifiro (approved):

PR #201, opened 2025-01-08
:alembic: Implement distributed adapter cache

joerunde:
## Description
vLLM currently does not support a distributed adapter cache- all replicas of a deployment must receive an explicit `/v1/load_lora_adapter` call to load an adapter.

This PR implements the existing `ADAPTER_CACHE` logic on top of vLLM's http server by injecting middleware that will detect if the `model` field of a request is referencing a set of files from the cache, and if so will pre-load the adapter before continuing with the call.

This also wraps the `/v1/models` endpoint to pre-load all adapters from the cache, so that the response is consistent across all replicas of a deployment.

Looking for some feedback here- I would like to solve this upstream but this gives us a quick way to roll out distributed lora adapters to users that matches existing TGIS behavior

## Merge criteria:

### DIFF
--- a/src/vllm_tgis_adapter/grpc/adapters.py
+++ b/src/vllm_tgis_adapter/grpc/adapters.py
@@ -49,7 +49,9 @@ class AdapterMetadata:
 class AdapterStore:
     cache_path: str  # Path to local store of adapters to load from
     adapters: dict[str, AdapterMetadata]
-    next_unique_id: int = 1
+    next_unique_id: int = (
+        262144  # Start at 2^18 to not run into vLLM's dynamic adapters
+    )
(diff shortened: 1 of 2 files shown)


dtrifiro (approved):
Looks good. Planning on adding tests?

joerunde:
@dtrifiro yeah, first planning on talking over with our team and seeing if this is a direction we want to go for short term deliverables vs. opening an RFC upstream in vLLM to implement there first. 

If we go with here, I'll add tests

joerunde:
Okay, team consensus is that this approach is reasonable, but we do not need to move fast to implement this in the adapter. I'll close this for now and use it as reference for a later RFC for vLLM

PR #203, opened 2025-01-08
nox: use uv as default venv backend

dtrifiro:
This should speed up venv creation considerably


dtrifiro:
Merge after #202

dtrifiro:
Docs updated as well in [1318463](https://github.com/opendatahub-io/vllm-tgis-adapter/commit/131846388470a74710bd712d4da3aa55a34a423f)

PR #206, opened 2025-01-13
:bug: fixup correlation-id log for grpc calls

joerunde:
This fixes a goof I made with the `x-correlation-id` grpc metadata header.
I added a unit test this time so I don't break it again 😉

### DIFF
--- a/src/vllm_tgis_adapter/grpc/grpc_server.py
+++ b/src/vllm_tgis_adapter/grpc/grpc_server.py
@@ -87,6 +87,7 @@ ADD_SPECIAL_TOKENS: bool = os.getenv("ADD_SPECIAL_TOKENS", "true").lower() not i
     "0",
     "false",
 )
+CORRELATION_ID_HEADER = "x-correlation-id"
 
 
 def with_default(value: _T, default: _T) -> _T:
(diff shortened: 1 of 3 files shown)


joerunde:
(Tests failed because of bug fixed by #207)

src/vllm_tgis_adapter/grpc/grpc_server.py
```diff
@@ -242,7 +243,7 @@ async def Generate(
             )
             is_tracing_enabled = await self.engine.is_tracing_enabled()
             headers = dict(context.invocation_metadata())
-            logs.set_correlation_id(request_id_i, headers.get("x-correlation_id"))
+            logs.set_correlation_id(request_id_i, headers.get(CORRELATION_ID_HEADER))
```
dtrifiro, line 246: :thinking:  `headers.get()` might return `None`, but `set_correlation_id` handles `str` only

joerunde: Yeah, mostly just a type hinting issue there but I updated to not touch the cache if there's no correlation id

joerunde:
unfortunately the automated precommit update caused some unrelated changes here

dtrifiro:
Rebased on #209

dtrifiro (approved):

PR #207, opened 2025-01-13
:bug: await init_app_state

joerunde:
## Description
I broke this recently by making a change to vllm's init that will ensure the engine loads all provided lora modules at startup. This broke `init_app_state` because it needed to be async to wait on the engine.

This backwards-compatible change will check if `init_app_state` needs to be awaited, and awaits it.

## Merge criteria:


tjohnson31415 (approved):
LGTM!

PR #230, opened 2025-02-18
build: remove grpcio-tools build time dependency

dtrifiro:
`grpcio-tools` is not available as a wheel for some platforms, making building `vllm-tgis-adapter` require building `grpcio-tools` from scratch.

Since this is not always possible (e.g. bazel not available on rhel), and is quite time consuming, we just include the generated code in the repo.

Note: this could (should?) also be added as a pre-commit hook, although this might slow down development a bit, so I left this as a gha step, for now.


R3hankhan123:
@dtrifiro , if we build grpcio-tools from scratch it will take alot of time as suggested in the pr, but for s390x we install the package with one env variable. Using this variable installation will take max 5 mins
`GRPC_PYTHON_BUILD_SYSTEM_OPENSSL=1 python3 -m pip install grpcio-tools --no-cache-dir`

Screenshot 2025-02-20 at 12 01 05 PM

**Note:** grpcio-tools took only 1-2 mins to build rest of time is taken by grpcio

R3hankhan123:
after this vlllm-tgis-adapter is getting installed successfully after on s390x

R3hankhan123:
@dtrifiro one more observation I have made for s390x is that during `pip install vllm_tgis_adapter` it takes the wheel that is already present on pypi and builds only grpcio and hftransfer

R3hankhan123:
These packages are being installed
```
STEP 5/8: RUN --mount=type=cache,target=/root/.cache/pip     --mount=type=cache,target=/root/.cache/uv     --mount=type=bind,from=build,source=/workspace/vllm,target=/workspace/vllm     HOME=/root uv pip install vllm-tgis-adapter==0.6.2
Using Python 3.12.5 environment at: /opt/vllm
Resolved 117 packages in 248ms
Uninstalled 2 packages in 3ms
warning: Failed to hardlink files; falling back to full copy. This may lead to degraded performance.
         If the cache and target directories are on different filesystems, hardlinking may not be supported.
         If this is intentional, set `export UV_LINK_MODE=copy` or use `--link-mode=copy` to suppress this warning.
Installed 21 packages in 358ms
 + accelerate==1.3.0
 + cachetools==5.5.2
 + deprecated==1.2.18
 + googleapis-common-protos==1.68.0
 + grpcio==1.70.0
 + grpcio-health-checking==1.70.0
 + grpcio-reflection==1.70.0
 + hf-transfer==0.1.8
 - importlib-metadata==8.6.1
 + importlib-metadata==8.5.0
 + opentelemetry-api==1.30.0
 + opentelemetry-exporter-otlp==1.30.0
 + opentelemetry-exporter-otlp-proto-common==1.30.0
 + opentelemetry-exporter-otlp-proto-grpc==1.30.0
 + opentelemetry-exporter-otlp-proto-http==1.30.0
 + opentelemetry-proto==1.30.0
 + opentelemetry-sdk==1.30.0
 + opentelemetry-semantic-conventions==0.51b0
 + opentelemetry-semantic-conventions-ai==0.4.2
 - prometheus-client==0.21.1
 + prometheus-client==0.21.0
 + vllm-tgis-adapter==0.6.2
 ```

dtrifiro:
@joerunde could you take a look? I don't see any potential issues with doing this apart from the annoyance of having to commit the generated `.pyi` in the repo and having to re-generate those every time `generation.proto` is updated.

dtrifiro:
Seems this was not required after all.

PR #232, opened 2025-02-19
Stop caching LoRA requests and use vllm server's cache

jberkhahn:
## Description
When the TGIS adapter starts the vllm api server, keep a reference to it around and pass it down into our gprc server. Then when we need to validate an adapter request, we can check if the vllm api server's cache contains a preexisting lora request that matches the incoming request. If it does, we can just drop the validation attempt on the floor. If it doesn't proceed as normal, except that when we validate the lora request, DONT add it to our cache.

### DIFF
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -19,7 +19,7 @@ from vllm.utils import FlexibleArgumentParser
 from vllm_tgis_adapter.tgis_utils.logs import add_logging_wrappers
 
 from .grpc import run_grpc_server
-from .http import run_http_server
+from .http import build_http_server, run_http_server
 from .logging import DEFAULT_LOGGER_NAME, init_logger
 from .tgis_utils.args import EnvVarArgumentParser, add_tgis_args, postprocess_tgis_args
 from .utils import check_for_failed_tasks, write_termination_log
@@ -43,15 +43,16 @@ async def start_servers(args: argparse.Namespace) -> None:
     async with build_async_engine_client(args) as engine:
         add_logging_wrappers(engine)
 
+        vllm_server = await build_http_server(args, engine)
         http_server_task = loop.create_task(
-            run_http_server(args, engine, sock),
+            run_http_server(args, vllm_server, sock),
             name="http_server",
         )
         # The http server task will catch interrupt signals for us
         tasks.append(http_server_task)
 
         grpc_server_task = loop.create_task(
-            run_grpc_server(args, engine),
+            run_grpc_server(args, engine, vllm_server),
             name="grpc_server",
         )
         tasks.append(grpc_server_task)
--- a/src/vllm_tgis_adapter/grpc/adapters.py
+++ b/src/vllm_tgis_adapter/grpc/adapters.py
@@ -15,7 +15,7 @@ import tempfile
 from pathlib import Path
 from typing import TYPE_CHECKING
 
-from vllm.lora.request import LoRARequest
+from vllm.entrypoints.openai.protocol import ErrorResponse
 from vllm.prompt_adapter.request import PromptAdapterRequest
 
 from vllm_tgis_adapter.logging import init_logger
@@ -23,12 +23,21 @@ from vllm_tgis_adapter.tgis_utils.convert_pt_to_prompt import convert_pt_to_peft
 
 from .validation import TGISValidationError
 
+try:
+    from vllm.entrypoints.openai.protocol import LoadLoRAAdapterRequest
+except ImportError:
+    from vllm.entrypoints.openai.protocol import (
+        LoadLoraAdapterRequest as LoadLoRAAdapterRequest,
+    )
+
 if TYPE_CHECKING:
     from vllm.entrypoints.grpc.pb.generation_pb2 import (
         BatchedGenerationRequest,
         BatchedTokenizeRequest,
         SingleGenerationRequest,
     )
+    from vllm.entrypoints.openai.serving_models import OpenAIServingModels
+    from vllm.lora.request import LoRARequest
 
 global_thread_pool = None  # used for loading adapter files from disk
 
@@ -49,7 +58,8 @@ class AdapterMetadata:
 class AdapterStore:
     cache_path: str  # Path to local store of adapters to load from
     adapters: dict[str, AdapterMetadata]
-    next_unique_id: int = 1
+    # Pick a large number to avoid colliding with vllm's adapter IDs
+    next_unique_id: int = 1000001
     load_locks: dict[str, asyncio.Lock] = dataclasses.field(default_factory=dict)
 
 
@@ -58,6 +68,7 @@ async def validate_adapters(
     | BatchedGenerationRequest
     | BatchedTokenizeRequest,
     adapter_store: AdapterStore | None,
+    vllm_model_handler: OpenAIServingModels,
 ) -> dict[str, LoRARequest | PromptAdapterRequest]:
     """Validate the adapters.
 
@@ -81,6 +92,12 @@ async def validate_adapters(
 
     # Guard against concurrent access for the same adapter
     async with adapter_store.load_locks.setdefault(adapter_id, asyncio.Lock()):
+        # Check VLLM server lora cache if this request matches an existing
+        # LoRA adapter
+        for existing_lora_request in vllm_model_handler.lora_requests:
+            if existing_lora_request.lora_name == adapter_id:
+                return {"lora_request": existing_lora_request}
+
         # If not already cached, we need to validate that files exist and
         # grab the type out of the adapter_config.json file
         if (adapter_metadata := adapter_store.adapters.get(adapter_id)) is None:
@@ -107,16 +124,19 @@ async def validate_adapters(
             )
 
             # Add to cache
+            # Query vllm's cache for lora requests
+            if adapter_metadata.adapter_type == "LORA":
+                lora_request = await _load_lora_adapter(
+                    request,
+                    adapter_id,
+                    adapter_metadata,
+                    vllm_model_handler,
+                )
+                return {"lora_request": lora_request}
+            # Use our cache for everything else
             adapter_store.adapters[adapter_id] = adapter_metadata
 
     # Build the proper vllm request object
-    if adapter_metadata.adapter_type == "LORA":
-        lora_request = LoRARequest(
-            lora_name=adapter_id,
-            lora_int_id=adapter_metadata.unique_id,
-            lora_path=adapter_metadata.full_path,
-        )
-        return {"lora_request": lora_request}
     if adapter_metadata.adapter_type == "PROMPT_TUNING":
         prompt_adapter_request = PromptAdapterRequest(
             prompt_adapter_id=adapter_metadata.unique_id,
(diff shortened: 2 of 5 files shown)


src/vllm_tgis_adapter/__main__.py
```diff
@@ -36,15 +36,16 @@ async def start_servers(args: argparse.Namespace) -> None:
     async with build_async_engine_client(args) as engine:
         add_logging_wrappers(engine)
 
+        vllm_server = build_http_server(args)
```
joerunde, line 39: There's a small chance for a race condition at startup here, since the fastapi app isn't actually initialized until `await init_app_state(...)` finishes, which is still inside of `run_http_server`.

Probably not the biggest deal, and I think if the readiness checks are using the http server then we won't get traffic until the http server is fully initialized, but it might be best to make `build_http_server` async and move `init_app_state` into it so we can be fully sure the state is initialized before any potential access by the grpc server

src/vllm_tgis_adapter/grpc/adapters.py
```diff
@@ -107,7 +107,10 @@ async def validate_adapters(
             )
 
             # Add to cache
-            adapter_store.adapters[adapter_id] = adapter_metadata
+            # We don't cache LoRA requests because the vllm api server
+            # maintains its own cache
+            if adapter_metadata.adapter_type != "LORA":
+                adapter_store.adapters[adapter_id] = adapter_metadata
```
joerunde, line 113: 🤔 

In the case where we:
- Checked and saw that the REST server does not have this LoRA request in its cache, and
- Validated that this file path points to a lora adapter

We actually need to give this information to the REST server so that it can be responsible for creating and caching the LoRA request. If we don't, then we still end up in the situation where we've created a request for the adapter from two separate places, so we'll end up with two separate adapters loaded from the same data into the gpu cache in the engine. (And if we just drop caching altogether, we'll end up with N copies for N requests 😱 ) 

`vllm_server.state.openai_serving_models.load_lora_adapter` should do the caching on the REST side, I think

src/vllm_tgis_adapter/grpc/adapters.py
```diff
@@ -107,7 +107,10 @@ async def validate_adapters(
             )
 
             # Add to cache
-            adapter_store.adapters[adapter_id] = adapter_metadata
+            # We don't cache LoRA requests because the vllm api server
```
joerunde, line 110: There's also a collision issue with `adapter_store.next_unique_id` and the unique id that the REST server will be using for cached adapters. Both servers initialize a counter at 0 and increment it for each adapter. In practice, if we stop creating or caching LoRA adapter requests here and we never enable prompt tuning, then we'll never have a collision. But I think we should go ahead and make some changes to ensure that it doesn't happen

tests/test_adapters.py
```diff
@@ -88,15 +88,30 @@ async def test_adapters_are_cached():
...
+        adapter_id=adapter_name,
     )
 
+    adapter_store = AdapterStore(cache_path=FIXTURES_DIR, adapters={})
+
+    adapters_1 = await validate_adapters(request, adapter_store=adapter_store)
+    adapters_2 = await validate_adapters(request, adapter_store=adapter_store)
+
+    # Metadata isn't cached in our cache
+    assert len(adapter_store.adapters) == 0
```
joerunde, line 107: Yes, though we do want to also ensure it's cached in the REST server's cache.

I don't have a high "care level" of how that's implemented- you could have `adapter_store.adapters` add the data from the REST cache so that these tests don't actually change, and instead we ensure that e.g.
```
assert adapter_store.adapters[adapter_name] == vllm_server.state.openai_serving_models.lora_requests[0]
```

Or, keep this as is and don't cache here and simply
```
assert vllm_server.state.openai_serving_models.lora_requests[0] == adapters_1
```

joerunde:
Looking good so far! Just a couple things to fix up

src/vllm_tgis_adapter/__main__.py
jberkhahn: so, doesn't this race condition already exist because the vllm http server is initialized in an async manner, so it's possible our grpc server comes up before it's ready anyways? if I make `build_http_server` async i dont see how that solves that problem? If we're getting a request to load a lora adapter and the server isn't ready yet, that request is going to fail anyways when it hits the load adapter endpoint on the vllm server. But I guess i can move this over to async and wait until it's init'd

joerunde: Yes, but currently I don't think the grpc server actually depends on the http server in any way. It only depends on the engine to run the requests, and the engine should be ready to go before we build the servers

src/vllm_tgis_adapter/grpc/adapters.py
jberkhahn: Oh, I didn't realize we were loading it on the GPU as well, I assumed that request got sent over to vllm at some point 🤔

src/vllm_tgis_adapter/grpc/adapters.py
jberkhahn: so I think I was a bit mistaken about the exact behavior of the tgis adapter - i thought those requests were getting forwarded to vllm at some point rather than being executed manually here. assuming I make it forward those requests to the vllm server, how is there potential for collision? I see how some id's will be "skipped", where our counter is incremented and then that doesn't eventually result in a request being issues to vllm, but how would the same counter be reused? is it possible the vllm http server is receiving requests independently of the adapter? If that's the case, don't we have to have some kind of way to share the incrementing of the counter that's atomic?

src/vllm_tgis_adapter/http.py
```diff
@@ -22,17 +22,26 @@
...
 
 
+def build_http_server(
+    args: argparse.Namespace,
+) -> FastAPI:
+    # builds the vllm api server so we can pass reference to it
+    # within the tgis adapter
+
+    app = build_app(args)
+    return app
```
joerunde, line 32: Since we're extracting the app now and making assumptions about its state, let's make this async and run `init_app_state` in here as well

src/vllm_tgis_adapter/grpc/adapters.py
```diff
@@ -126,12 +145,36 @@ async def validate_adapters(
...
+    load_request = LoadLoRAAdapterRequest(
+        lora_path=adapter_metadata.full_path,
+        lora_name=adapter_id,
+    )
+    load_result = await vllm_model_handler.load_lora_adapter(
+        request=load_request,
+        base_model_name=request.model_id,
+    )
+    if isinstance(load_result, ErrorResponse):
+        raise TypeError(load_result.message)
```
joerunde, line 171: If we make this a `ValueError` instead then the exception handling in the server will ensure that the user gets a `StatusCode.INVALID_ARGUMENT` type response with this message, which would be helpful for them to know what went wrong

tests/test_adapters.py
```diff
@@ -57,66 +75,145 @@ async def test_prompt_adapter():
...
+            lora_name=adapter_name,
+        ),
+        base_model_name=base_model_name,
+    )
 
 
 @pytest.mark.asyncio
-async def test_adapters_are_cached():
-    adapter_name = "granite-3b-code-instruct-lora"
+async def test_adapters_are_cached(vllm_model_handler):
```
joerunde, line 120: maybe rename `test_prompt_adapters_are_cached_locally`?

tests/test_adapters.py
```diff
@@ -57,66 +75,145 @@ async def test_prompt_adapter():
...
+            [
+                LoRARequest(
+                    lora_name=adapter_name,
+                    lora_int_id=0,
+                    lora_path=lora_path,
+                ),
+            ],
+        ]
+    )
+    type(vllm_model_handler).lora_requests = mock_lora_requests
```
joerunde, line 98: `type(vllm_model_handler)` should resolve to the class `AsyncMock` so I think this will set `.lora_requests` for every instance of an AsyncMock from here on out...? Not sure that's quite what we want.

Is the issue that when you do `vllm_model_handler.lora_requests` it expects that to be an awaitable?

tests/test_adapters.py
```diff
@@ -57,66 +75,145 @@ async def test_prompt_adapter():
...
+async def test_lora_adapter(vllm_model_handler):
+    base_model_name = "/granite/granite-3b-base-v2/step_75000_ckpt"
     adapter_name = "granite-3b-code-instruct-lora"
+    lora_path = str(FIXTURES_DIR) + "/" + adapter_name
     request = BatchedGenerationRequest(
         adapter_id=adapter_name,
+        model_id=base_model_name,
     )
+    mock_lora_requests = PropertyMock(
+        side_effect=[
```
joerunde, line 87: I expected this to be `return_value` instead of `side_effect` 🤔🤔🤔

(But not a problem if this works)

tests/test_adapters.py
```diff
@@ -10,19 +11,34 @@
...
+        LoadLoraAdapterRequest as LoadLoRAAdapterRequest,
+    )
+
 FIXTURES_DIR = Path(__file__).parent / "fixtures"
 
 
+@pytest.fixture
+def vllm_model_handler() -> AsyncMock:
+    mock_handler = AsyncMock()
+    return mock_handler
```
joerunde, line 27: I wonder if for this test it would be easier to write a class like this:
```
class FakeModelsHandler:

    def __init__(self):
        self.lora_requests = []

    async def load_lora_adapter(self, request):
        if request.lora_name.starts_with("poison"):
            raise ValueError("oops!")
        self.lora_requests.append( #make a LoRARequest from the LoadLoRAAdapterRequest)
```

Seems simpler than dealing with the mocks since we're only using this as a simple cache

tests/test_adapters.py
```diff
@@ -57,66 +75,145 @@ async def test_prompt_adapter():
...
     assert (
-        adapters_1["lora_request"].lora_int_id
+        adapters_1["prompt_adapter_request"].prompt_adapter_id
         < adapters_2["prompt_adapter_request"].prompt_adapter_id
     )
 
 
 @pytest.mark.asyncio
-async def test_cache_handles_concurrent_loads():
+async def test_cache_handles_concurrent_loads(vllm_model_handler):
```
joerunde, line 216: For completeness we could do this with lora adapters too just to make sure our locking works so that we only load once

src/vllm_tgis_adapter/grpc/adapters.py
jberkhahn: so it was originally a ValueError, but our linter was complaining and wanted to make it a value error instead, how do i make an exception for the linter?

tests/test_adapters.py
jberkhahn: i'm not 100% clear on what the mock is doing here, but when i just assign `vllm_model_handler.lora_requests" to the PropertyMock, I have to call `vllm_model_handler.lora_requests()` to get the expected value out of it, when I want to reference it as an attribute `vllm_model_handler.lora_requests`.

tests/test_adapters.py
jberkhahn: so the name on this is really confusing. return value (to my understanding) only lets you specify a single return value which is always the same. for this test to work I need to be able to set multiple different return values in sequence to fake the expected behavior of vllm actually loading and caching the adapter. side_effect seems to just be a way that lets you specify multiple return values that supersede return_value even if it's set, which to me doesn't seem like a side effect but more of a primary effect? The naming me threw me for a loop when I was trying to figure this out.

src/vllm_tgis_adapter/grpc/adapters.py
joerunde: The rules for ruff can all be found here: https://docs.astral.sh/ruff/rules/

They can be turned off individually by adding a comment at the end of the line like `## noqa: ${code}`

In this case you can see the error is coming from [tryceratops](https://docs.astral.sh/ruff/rules/#tryceratops-try) and you can disable with `## noqa: TRY004`

tests/test_adapters.py
joerunde: ah, yeah that sounds about right for python mocks!

tests/test_adapters.py
joerunde: I looked into it a bit more, and it looks to me like the `type(vllm_model_handler)` behavior actually returns a different instance of a class for each mock, e.g.
```
print( type(Mock()) == type(Mock()) )
```
prints `False`. So I think this actually is safe to do, it's just... pretty interesting. But I guess that's how it goes for python mocks

joerunde (approved):
Nice!

PR #224, opened 2025-02-16
Pass param for pre-created socket to uvicorn

fialhocoelho:
During the integration tests, the built image triggered the error below:
```
[rank0]:[W215 15:13:14.568778632 ProcessGroupNCCL.cpp:1496] Warning: WARNING: destroy_process_group() was not called before program exit, which can leak resources. For more info, please see https://pytorch.org/docs/stable/distributed.html#shutdown (function operator())
Traceback (most recent call last):
  File "/opt/vllm/lib64/python3.12/site-packages/vllm_tgis_adapter/http.py", line 66, in run_http_server
    shutdown_coro = await serve_http(app, **serve_kwargs)
                          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: serve_http() missing 1 required positional argument: 'sock'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/opt/vllm/lib64/python3.12/site-packages/vllm_tgis_adapter/__main__.py", line 114, in <module>
    run_and_catch_termination_cause(loop, task)
  File "/opt/vllm/lib64/python3.12/site-packages/vllm_tgis_adapter/__main__.py", line 90, in run_and_catch_termination_cause
    loop.run_until_complete(task)
  File "uvloop/loop.pyx", line 1518, in uvloop.loop.Loop.run_until_complete
  File "/opt/vllm/lib64/python3.12/site-packages/vllm_tgis_adapter/__main__.py", line 80, in start_servers
    raise RuntimeError(f"Failed task={name} ({coro_name})") from exception
RuntimeError: Failed task=http_server (run_http_server)
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1739632399.252718       1 init.cc:232] grpc_wait_for_shutdown_with_timeout() timed out.
```

Looking for the function `serve_http()` in `vllm/entrypoints/launcher.py`, I found that a recent modification in PR [#13113](https://github.com/vllm-project/vllm/pull/13113) (post vLLM v0.7.2) added an optional parameter `sock` to allow uvicorn to use a pre-created socket.

We actually use the `serve_http()` function in `/src/vllm_tgis_adapter/http.py`. Because of this, I added a parameter `sock=None` to maintain the same function signature we used before. I'm not sure if we need to use a pre-created socket for multi-GPU support or for something else.

PR #233, opened 2025-02-21
create socket before starting the engine and pass socket into the vllm server on startup

jberkhahn:
VLLM made a breaking change recently here: https://github.com/vllm-project/vllm/pull/13113

But it looks like they've been doing stuff around passing that socket in for a while to avoid some stuff like race conditions? Figured I'd port over the functionality and get it on vllm version.

### DIFF
--- a/src/vllm_tgis_adapter/__main__.py
+++ b/src/vllm_tgis_adapter/__main__.py
@@ -11,6 +11,7 @@ import uvloop
 import vllm
 from vllm.entrypoints.openai.api_server import (
     build_async_engine_client,
+    create_server_socket,
 )
 from vllm.entrypoints.openai.cli_args import make_arg_parser
 from vllm.utils import FlexibleArgumentParser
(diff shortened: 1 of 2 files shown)


src/vllm_tgis_adapter/http.py
```diff
@@ -62,6 +64,10 @@ async def set_correlation_id(request: Request, call_next: Callable) -> Response:
         "ssl_cert_reqs": args.ssl_cert_reqs,
     }
     serve_kwargs.update(uvicorn_kwargs)
+    from vllm.version import __version_tuple__ as vllm_version_tuple
+    ...
```
joerunde, line 68: why the `...` here?

src/vllm_tgis_adapter/__main__.py
```diff
@@ -28,16 +29,33 @@
 
 logger = init_logger(DEFAULT_LOGGER_NAME)
 
+def create_server_socket(addr: Tuple[str, int]) -> socket.socket:
```
joerunde, line 32: can we import this directly from vllm instead of copy-pasting?

src/vllm_tgis_adapter/http.py
jberkhahn: was copying from jeff's code here: https://github.com/opendatahub-io/vllm-tgis-adapter/pull/224/files

joerunde: Right, I think that's probably copy-pasted from an example where the `...` was meant to elide the code that should go in between. A literal here `...` is a value in python of type `ellipsis`, which we don't need hanging out here.

The import should go at the top of the file, and then the tuple should be checked here

src/vllm_tgis_adapter/__main__.py
```diff
@@ -3,17 +3,19 @@
 import asyncio
 import contextlib
 import os
+import socket
 import traceback
 from concurrent.futures import FIRST_COMPLETED
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Tuple
```
joerunde, line 9: Tuple should be unused now, but the linter should catch this

jberkhahn: went back and ran the linter

dtrifiro (approved):

dtrifiro:
Thanks, looks good.

PR #241, opened 2025-03-31
gha: fix ci caching

dtrifiro:
- python version cache key was referring to an old (now unset) variable
- add vllm version as nox envs cache key
- cache huggingface hub directory, this should prevent getting temporary 403 errors from huggingface, which has made CI flakey lately

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -78,6 +78,13 @@ jobs:
           path: /home/runner/.nox
           key: ${{ runner.os }}-python-${{matrix.pyv}}-${{ hashFiles('noxfile.py') }}-${{matrix.vllm_version}}
 
+      - name: hf hub cache
+        id: hf-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface/hub
+          key: ${{ runner.os }}
+
       - name: Lint code and check dependencies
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -76,7 +77,14 @@ jobs:
...
           path: /home/runner/.nox
-          key: ${{ runner.os }}-python-${{steps.setup-python.outputs.python-version }}-${{ hashFiles('noxfile.py') }}
+          key: ${{ runner.os }}-python-${{matrix.pyv}}-${{ hashFiles('noxfile.py') }}-${{matrix.vllm_version}}
+
+      - name: hf hub cache
+        id: hf-cache
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/huggingface/hub
+          key: ${{ runner.os }}
```
tjohnson31415, line 87: Is the HF Cache specific to the OS or can it be shared across OSes?

dtrifiro: I believe it can be shared across OSes, but I just needed a dummy key because the cache complains otherwise. I don't think we're even interested in running this workflow for multiple OSes.

tjohnson31415 (approved):
LGTM

dtrifiro:
Thanks @tjohnson31415, merging this as is for now, CI failures are related to the model dtype dtype being set to half, leaving those fixes  to another PR

PR #247, opened 2025-04-15
:recycle: Consolidate vllm utils and allow vllm version to be overwritten

evaline-ju:
## Description
- PR `https://github.com/vllm-project/vllm/pull/16533` will remove the import of `StoreBoolean` through `vllm.engine.arg_utils`, but since this is done by an import from `vllm.utils` anyway, the import can be directly used
- It appeared the wrong vllm versions were being "tested" during the tests - this has been updated
- PR `https://github.com/vllm-project/vllm/pull/16572` will change the inputs to `init_app_state` from `model_config` to `vllm_config`

## How Has This Been Tested?

Mainly discovered through another adapter and tested with previous versions of vllm (0.8.1-0.8.2, discovered the problem with latest main)

For the vllm version addition (see #248) : 
Other dependencies in `.[tests]` appear to be changing the vllm version itself from the initial override - a suspected example is something like `https://github.com/vllm-project/vllm/pull/15224` where the opentelemetry dependencies are pinned lower than what this adapter currently allows.

However, swapping the `.[tests]` and vllm version dependencies entirely also led to unexpected behavior such as overriding protobuf versions, leading to more issues. The compromise then appeared to be re-installing the expected vllm version itself without dependencies

## Merge criteria:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -23,6 +23,11 @@ def tests(session: nox.Session) -> None:
 
     session.install(".[tests]")
 
+    # Re-install without dependencies since `.[tests]` brings in
+    # overrides that may bring the vllm version down
+    if vllm_version := os.getenv("VLLM_VERSION_OVERRIDE"):
+        session.install(vllm_version, "--no-deps")
+
     session.run(
         "pytest",
         "--cov",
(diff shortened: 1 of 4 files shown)


joerunde (approved):
LGTM assuming it all passes

dtrifiro:
Failures on vllm@main are unrelated, main is failing as well.

joerunde (approved):
lgtm

PR #271, opened 2025-08-09
unpin vllm for rocm

heyselbi:
## Description
ROCM Build fails with
```
[8/8] STEP 4/8: RUN --mount=type=cache,target=/root/.cache/pip     --mount=type=cache,target=/root/.cache/uv     --mount=type=bind,from=build_vllm,src=/workspace/dist,target=/install/vllm/     HOME=/root uv pip install         "$(echo /install/vllm/*.whl)[audio,video,tensorizer]"         vllm-tgis-adapter==${VLLM_TGIS_ADAPTER_VERSION}
Using Python 3.12.9 environment at: /opt/vllm
  × No solution found when resolving dependencies:
  ╰─▶ Because only vllm[audio]==0.1.dev8486+ga2a0e2a.rocm634 is available
      and vllm-tgis-adapter==0.8.0 depends on vllm>=0.10.0, we can conclude
      that vllm-tgis-adapter==0.8.0 and all versions of vllm[audio] are
      incompatible.
      And because you require vllm-tgis-adapter==0.8.0 and vllm[audio], we can
      conclude that your requirements are unsatisfiable.
```

## Merge criteria:

## Summary by CodeRabbit

* **Chores**
  * Updated the "vllm" dependency to remove the minimum version requirement.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 requires-python = ">=3.9"
 dynamic = ["version"]
 dependencies = [
-  "vllm>=0.10.0",
+  "vllm",
   "prometheus_client==0.21.1",
   "grpcio==1.70.0",
   "grpcio-health-checking==1.70.0",


dtrifiro:
We need at least `vllm>=0.7.2`

PR #278, opened 2025-09-18
bump minimum vllm version to >=0.10.2

dtrifiro:
* vllm>=0.10.2
* remove guided decoding (removed in v0.10.2)


dtrifiro:
removed in favour of https://github.com/opendatahub-io/vllm-tgis-adapter/pull/286

PR #281, opened 2025-09-25
fix: fix lora breakage in vLLM > 0.10.0

njhill:
`lora_requests` changed from list to dict in https://github.com/vllm-project/vllm/pull/20636.

This change also retains backwards compatibility with older versions.

## Summary by CodeRabbit

* **Bug Fixes**
  * Better compatibility across vLLM versions for LoRA adapters, reducing version-specific errors.
  * Reuses cached LoRA adapter requests to avoid redundant loads and reduce failures.

* **Refactor**
  * Made adapter validation and loading version-agnostic and more robust while preserving existing behavior.

### DIFF
--- a/src/vllm_tgis_adapter/grpc/adapters.py
+++ b/src/vllm_tgis_adapter/grpc/adapters.py
@@ -91,9 +91,10 @@ async def validate_adapters(
     async with adapter_store.load_locks.setdefault(adapter_id, asyncio.Lock()):
         # Check VLLM server lora cache if this request matches an existing
         # LoRA adapter
-        for existing_lora_request in vllm_model_handler.lora_requests:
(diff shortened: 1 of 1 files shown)


gkumbhat:
Hi @njhill 

vLLM `0.10.1` has removed the guided decoding module right? So [this](https://github.com/opendatahub-io/vllm-tgis-adapter/blob/main/src/vllm_tgis_adapter/tgis_utils/guided_decoding.py) will start to break with vLLM update post `0.10.1` ?

njhill:
Hey @gkumbhat, yes it looks like additional compatibility updates will be needed for that. Guided decoding is still supported (we are usually calling it structured output these days), so hopefully it should just be a matter of connecting properly to the updated codepath.

PR #283, opened 2025-09-25
fix vllm build

dtrifiro:
The vllm isolated build for CPU is currently broken resulting in some `torch`-related not being found in the `_C.abi3.so` during testing, breaking CI

This PR addresses this by retrieving the build dependencies beforehand (when `VLLM_VERSION_OVERRIDE` is defined) and pre-installing them, to allow for vllm to be built using `--no-build-isolation`

This only works with local paths and `git+https://github.com/vllm-project/vllm@ref` style overrides (such as those used in CI)

## Summary by CodeRabbit

- Chores
  - Centralized, idempotent handling for optional vLLM wheel overrides; CI can check out and build specified vLLM refs and provide them as an override.
  - Added a script to install vLLM build dependencies and support building/providing vLLM wheels.
  - Simplified CI matrix/caching and lint setup; added a lint ignore for the new helper script.
- Documentation
  - Updated installation guidance to a from-source vLLM build-and-wheel workflow.
- Tests
  - vLLM override installation invoked consistently before/after deps; test CLI args adjusted and a parameterized test case removed.
  - Increased health-check patience and improved timeout/error reporting in test utilities.

### DIFF
--- a/.github/scripts/install_vllm_build_deps.py
+++ b/.github/scripts/install_vllm_build_deps.py
@@ -1,0 +1,62 @@
+#!/usr/bin/env -S uv run --active --script
+# /// script
+# dependencies = [
+#   "tomli",
+# ]
+# ///
+from __future__ import annotations
+
+import sys
+from pathlib import Path
+from subprocess import run
+
+import tomli
+
+
+def _get_build_dependencies_from_file(pyproject: Path) -> list[str]:
+    if not pyproject.exists():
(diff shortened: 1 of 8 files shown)


heyselbi (approved):

dtrifiro:
Tests for `vllm==0.10.0` are finally passing. Running lora tests on cpu is still failing, but I'll leave that for another PR.

